PR #141, opened 2025-01-16
Enhance MongoDB Connection Management with Reconnection Logic

gifflet:
This PR improves the MongoDB connection management by introducing a reconnection mechanism and enhancing error handling:

**Reconnection Logic:**

- Added reconnectMongo function to handle MongoDB reconnection attempts if the connection is lost.
- Automatically retries reconnections with detailed logs for each attempt.
- Resets the error counter when the connection is successfully restored.

### DIFF
--- a/common/db/db.go
+++ b/common/db/db.go
@@ -85,15 +85,10 @@ func checkDBConnection() {
 					"error":       err,
 				}).Error("Failed to reconnect to MongoDB")
 
-				if errorCounter >= 10 {
-					log.Fatal("Lost connection to MongoDB server and failed to reconnect after 10 attempts!")
+				if errorCounter >= 30 {
+					log.Fatal("Lost connection to MongoDB server and failed to reconnect after 30 attempts!")
 				}
 			} else {
-				log.Info("MongoDB connection restored")
-				errorCounter = 0
-			}
-		} else {
-			if errorCounter > 0 {
 				log.Info("MongoDB connection restored")
 				errorCounter = 0
 			}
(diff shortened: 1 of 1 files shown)


common/db/db.go
```diff
@@ -66,18 +69,57 @@ func CloseMongoConn() {
...
+				log.Info("MongoDB connection restored")
+				errorCounter = 0
 			}
 		} else {
-			log.Fatal("Lost connection to MongoDB server for more than 10 seconds!")
+			if errorCounter > 0 {
+				log.Info("MongoDB connection restored")
+				errorCounter = 0
+			}
+		}
```
shamanec, line 100: I am not sure this else statement is actually needed, you already set `errorCounter` to 0 on successful reconnect, when this block is reached it should already be in healthy state

common/db/db.go
```diff
@@ -66,18 +69,57 @@ func CloseMongoConn() {
...
+				"error":       err,
+			}).Warn("Failed to ping MongoDB")
+
+			if err := reconnectMongo(); err != nil {
+				log.WithFields(log.Fields{
+					"error_count": errorCounter,
+					"error":       err,
+				}).Error("Failed to reconnect to MongoDB")
+
+				if errorCounter >= 10 {
```
shamanec, line 88: Would you mind increasing this to 30? I've been thinking for a while that this is too strict, you should also change the log message below along with it

shamanec (changes requested):
just two small comments

common/db/db.go
gifflet: The else block was removed in the last commit

shamanec (approved):

shamanec:
Thank you for the totally unexpected contribution ❤️

PR #145, opened 2025-01-28
Implement Per-Device Stream Settings with Global Management

gifflet:
This PR introduces enhanced management for stream settings, allowing global configurations as well as per-device customizations:

- Global Stream Settings:
  - Extended StreamSettings to support platform-specific scaling factors (ScalingFactorAndroid and ScalingFactoriOS).
  - Added GlobalSettings tab in the UI (GlobalSettings.js) to manage global configurations via a dedicated admin panel.
  - New API endpoints (GET and POST) to retrieve and update global settings.

- Per-Device Stream Settings:
  - Added DeviceStreamSettings model and database methods (GetDeviceStreamSettings and UpdateDeviceStreamSettings)  for storing and retrieving custom stream settings.
  - Modified setupDevices logic to prioritize per-device settings, with fallback to global configurations.

- Periodic Update Routine:
  - Introduced UpdateDevicesStreamSettings to periodically sync stream settings across devices, ensuring consistency in live states.

### DIFF
--- a/common/db/db.go
+++ b/common/db/db.go
@@ -491,3 +491,88 @@ func UploadFileGridFS(file io.Reader, fileName string, force bool) error {
 		return nil
 	}
 }
+
+func GetGlobalStreamSettings() (models.StreamSettings, error) {
+	var globalSettings models.GlobalSettings
+	var streamSettings models.StreamSettings
+
+	coll := mongoClient.Database("gads").Collection("global_settings")
+	filter := bson.D{{Key: "type", Value: "stream-settings"}}
+
+	err := coll.FindOne(mongoClientCtx, filter).Decode(&globalSettings)
+	if err == mongo.ErrNoDocuments {
+		streamSettings = models.StreamSettings{
+			TargetFPS:            15,
+			JpegQuality:          75,
+			ScalingFactorAndroid: 50,
+			ScalingFactoriOS:     50,
+		}
+
+		err = UpdateGlobalStreamSettings(streamSettings)
+		if err != nil {
+			return streamSettings, err
+		}
+	} else if err != nil {
+		return streamSettings, err
+	} else {
+		settingsBytes, err := bson.Marshal(globalSettings.Settings)
+		if err != nil {
+			return streamSettings, fmt.Errorf("failed to marshal settings: %v", err)
+		}
+
+		err = bson.Unmarshal(settingsBytes, &streamSettings)
+		if err != nil {
+			return streamSettings, fmt.Errorf("failed to unmarshal settings: %v", err)
+		}
+	}
+
+	return streamSettings, nil
+}
+
+func UpdateGlobalStreamSettings(settings models.StreamSettings) error {
+	globalSettings := models.GlobalSettings{
+		Type:        "stream-settings",
+		Settings:    settings,
+		LastUpdated: time.Now(),
+	}
+
+	update := bson.M{
+		"$set": globalSettings,
+	}
+
+	coll := mongoClient.Database("gads").Collection("global_settings")
+	filter := bson.D{{Key: "type", Value: "stream-settings"}}
+	opts := options.Update().SetUpsert(true)
+
+	_, err := coll.UpdateOne(mongoClientCtx, filter, update, opts)
+	return err
+}
+
+func GetDeviceStreamSettings(udid string) (models.DeviceStreamSettings, error) {
+	var deviceStreamSettings models.DeviceStreamSettings
+
+	coll := mongoClient.Database("gads").Collection("device_stream_settings")
+	filter := bson.D{{Key: "udid", Value: udid}}
+
+	err := coll.FindOne(mongoClientCtx, filter).Decode(&deviceStreamSettings)
+	if err != nil {
+		if err == mongo.ErrNoDocuments {
+			return deviceStreamSettings, nil // No settings found, return empty struct
+		}
+		return deviceStreamSettings, err // Return error if any other error occurred
+	}
+
+	return deviceStreamSettings, nil
+}
+
+func UpdateDeviceStreamSettings(udid string, settings models.DeviceStreamSettings) error {
+	coll := mongoClient.Database("gads").Collection("device_stream_settings")
+	filter := bson.D{{Key: "udid", Value: udid}}
+	update := bson.M{
+		"$set": settings,
+	}
+	opts := options.Update().SetUpsert(true)
+
+	_, err := coll.UpdateOne(mongoClientCtx, filter, update, opts)
+	return err
+}
--- a/common/db/db.go
+++ b/common/db/db.go
@@ -556,10 +556,7 @@ func GetDeviceStreamSettings(udid string) (models.DeviceStreamSettings, error) {
 
 	err := coll.FindOne(mongoClientCtx, filter).Decode(&deviceStreamSettings)
 	if err != nil {
-		if err == mongo.ErrNoDocuments {
-			return deviceStreamSettings, nil // No settings found, return empty struct
-		}
-		return deviceStreamSettings, err // Return error if any other error occurred
+		return deviceStreamSettings, err // Return error if no settings were found or if any other error occurred
 	}
 
 	return deviceStreamSettings, nil
--- a/common/sync.go
+++ b/common/sync.go
@@ -1,0 +1,5 @@
+package common
+
+import "sync"
+
+var Mu sync.Mutex
--- a/hub/gads-ui/src/components/Admin/GlobalSettings/GlobalSettings.js
+++ b/hub/gads-ui/src/components/Admin/GlobalSettings/GlobalSettings.js
@@ -1,0 +1,158 @@
+import { Box, Button, TextField, Grid, MenuItem, Stack, Typography } from '@mui/material';
+import { useState, useEffect } from 'react';
+import { api } from '../../../services/api';
+
+// Import options from StreamSettings
+const fpsOptions = [5, 10, 15, 20, 30, 45, 60];
+const jpegQualityOptions = [25, 30, 35, 40, 45, 50, 55, 60, 65, 70, 75, 80, 85, 90];
+const scalingFactorOptionsAndroid = [25, 50];
+const scalingFactorOptionsiOS = [25, 30, 35, 40, 45, 50, 55, 60, 65, 70, 75, 80, 85, 90, 95, 100];
+
+export default function GlobalSettings() {
+    const [fps, setFps] = useState(15);
+    const [jpegQuality, setJpegQuality] = useState(75);
+    const [scalingFactorAndroid, setScalingFactorAndroid] = useState(50);
+    const [scalingFactoriOS, setScalingFactoriOS] = useState(50);
+    const [loading, setLoading] = useState(false);
+    const [status, setStatus] = useState(null);
+
+    useEffect(() => {
+        // Fetch global stream settings on component mount
+        api.get('/admin/global-settings')
+            .then(response => {
+                const settings = response.data;
+                setFps(settings.target_fps);
+                setJpegQuality(settings.jpeg_quality);
+                setScalingFactorAndroid(settings.scaling_factor_android);
+                setScalingFactoriOS(settings.scaling_factor_ios);
+            })
+            .catch(() => {
+                setStatus('error');
+            });
+    }, []);
+
+    const handleSave = () => {
+        setLoading(true);
+        const settings = { target_fps: fps, jpeg_quality: jpegQuality, scaling_factor_android: scalingFactorAndroid, scaling_factor_ios: scalingFactoriOS };
+
+        api.post('/admin/global-settings', settings)
+            .then(() => {
+                setStatus('success');
+            })
+            .catch(() => {
+                setStatus('error');
+            })
+            .finally(() => {
+                setLoading(false);
+                setTimeout(() => setStatus(null), 2000);
+            });
+    };
+
+    return (
+        <Stack direction='row' spacing={2} style={{ width: '100%', marginLeft: '10px', marginTop: '10px' }}>
+            <Box style={{
+                marginBottom: '10px',
+                height: '80vh',
+                overflowY: 'scroll',
+                border: '2px solid black',
+                borderRadius: '10px',
+                boxShadow: 'inset 0 -10px 10px -10px #000000',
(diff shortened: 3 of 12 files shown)


common/sync.go
```diff
@@ -0,0 +1,5 @@
+package common
+
+import "sync"
+
+var Mu sync.Mutex
```
shamanec, line 5: If we are going to have mutexes in common it might be better to have target mutexes - for example in the current situation a mutex for the stream settings. If we reuse the same mutex across the whole app we could be locking resources unrelated to each other. What do you think about having a MutexManager? Something like:
`
type MutexManager struct {
    StreamSettings sync.Mutex
}
`
and initialize and use that

common/db/db.go
```diff
@@ -491,3 +491,88 @@ func UploadFileGridFS(file io.Reader, fileName string, force bool) error {
...
+func GetDeviceStreamSettings(udid string) (models.DeviceStreamSettings, error) {
+	var deviceStreamSettings models.DeviceStreamSettings
+
+	coll := mongoClient.Database("gads").Collection("device_stream_settings")
+	filter := bson.D{{Key: "udid", Value: udid}}
+
+	err := coll.FindOne(mongoClientCtx, filter).Decode(&deviceStreamSettings)
+	if err != nil {
+		if err == mongo.ErrNoDocuments {
+			return deviceStreamSettings, nil // No settings found, return empty struct
```
shamanec, line 560: Why don't we just return the error in both cases? Then we can avoid the unnecessary check for device UDID in `provider/devices/common.go` and just fallback to the default stream settings

provider/devices/common.go
```diff
@@ -108,9 +112,28 @@ func setupDevices() {
...
+
+		// Get the DeviceStreamSettings for the current device
+		deviceStreamSettings, err := db.GetDeviceStreamSettings(dbDevice.UDID)
+		if err != nil {
+			logger.ProviderLogger.LogError("setupDevices", fmt.Sprintf("Failed to retrieve stream settings for device `%s`: %v", dbDevice.UDID, err))
+			continue
+		}
+
+		// Check if the DeviceStreamSettings exist
+		if deviceStreamSettings.UDID == "" {
```
shamanec, line 124: Addition to [this](https://github.com/shamanec/GADS/pull/145/files#r1933758086). If we return an error when the there is no document directly then we can skip this check and directly fallback and do `updateDeviceWithGlobalSettings` above

provider/devices/common.go
```diff
@@ -108,9 +112,28 @@ func setupDevices() {
...
-		dbDevice.StreamTargetFPS = 15
-		dbDevice.StreamJpegQuality = 75
-		dbDevice.StreamScalingFactor = 50
+
+		// Get the DeviceStreamSettings for the current device
+		deviceStreamSettings, err := db.GetDeviceStreamSettings(dbDevice.UDID)
+		if err != nil {
+			logger.ProviderLogger.LogError("setupDevices", fmt.Sprintf("Failed to retrieve stream settings for device `%s`: %v", dbDevice.UDID, err))
+			continue
+		}
```
shamanec, line 121: If we directly return an error when there is no document for the device as stated below`
deviceStreamSettings, err := db.GetDeviceStreamSettings(dbDevice.UDID)
		if err != nil {
			logger.ProviderLogger.LogError("setupDevices", fmt.Sprintf("Failed to retrieve stream settings for device `%s`: %v", dbDevice.UDID, err))
						err = updateDeviceWithGlobalSettings(dbDevice)
			if err != nil {
				logger.ProviderLogger.LogError("setupDevices", fmt.Sprintf("Failed to update device `%s` with global settings: %v", dbDevice.UDID, err))
				continue
			}
		} else {
                      // Apply the retrieved stream settings
			dbDevice.StreamTargetFPS = deviceStreamSettings.StreamTargetFPS
			dbDevice.StreamJpegQuality = deviceStreamSettings.StreamJpegQuality
			dbDevice.StreamScalingFactor = deviceStreamSettings.StreamScalingFactor
}

`

hub/gads-ui/src/components/Admin/GlobalSettings/GlobalSettings.js
```diff
@@ -0,0 +1,158 @@
...
+                                    disabled={loading}
+                                    style={{
+                                        backgroundColor: '#2f3b26',
+                                        color: '#f4e6cd',
+                                        fontWeight: 'bold'
+                                    }}
+                                >
+                                    {loading ? 'Saving...' : 'Save Settings'}
+                                </Button>
+                                {status === 'success' && <Typography color="green">Settings saved successfully!</Typography>}
```
shamanec, line 149: Would you mind using the global snackbar like [here](https://github.com/shamanec/GADS/blob/3db8d6d9e2ba976048758b611d39c74c80c85b4d/hub/gads-ui/src/components/DeviceControl/StreamCanvas/StreamSettings.js#L60). I am trying to move to similar feedback across the app

provider/devices/common.go
```diff
@@ -908,3 +931,68 @@ func checkAppiumUp(device *models.Device) {
...
+	return nil
+}
+
+func UpdateDevicesStreamSettings() {
+	ticker := time.NewTicker(1 * time.Second)
+	defer ticker.Stop()
+
+	for range ticker.C {
+		for _, dbDevice := range DBDeviceMap {
+			time.Sleep(1 * time.Second)
```
shamanec, line 960: I think this sleep should be outside the DBDeviceMap loop, why are we waiting per device instead of per ticker?

common/sync.go
gifflet: Well... the way I implemented the common mutex could really cause us some issues in the future. So, your suggestion is to define the MutexManager (renaming the sync.go to mutex_manager.go) and use its mutexes in the whole application, right?

provider/devices/common.go
```diff
@@ -908,3 +931,68 @@ func checkAppiumUp(device *models.Device) {
...
+	if dbDevice.OS == "android" {
+		dbDevice.StreamScalingFactor = globalSettings.ScalingFactorAndroid
+	} else if dbDevice.OS == "ios" {
+		dbDevice.StreamScalingFactor = globalSettings.ScalingFactoriOS
+	}
+
+	return nil
+}
+
+func UpdateDevicesStreamSettings() {
```
shamanec, line 954: I am not sure this goroutine does anything. Even if you update the global stream settings, the actual update on the currently provisioned device happens only with the `update-stream-settings` API call. So even if this for some reason changes the values in the DBDeviceMap they will not be reflected on a device that is already live - it will have to be re-provisioned - and if it is re-provisioned it will take the new global stream settings into account either way

common/sync.go
gifflet: The refactor is available [in this commit](https://github.com/shamanec/GADS/pull/145/commits/8ada1eebe18344f4cf6e0d465c0e8f7b7057257a)

shamanec: Yeah, I guess having them centralized would be better in the cases it makes sense 😅 now there are tons of them here and there

common/db/db.go
gifflet: Ok, it makes sense. I will work on it

provider/devices/common.go
gifflet: I think this was already adjusted in the previous review ([commit](https://github.com/shamanec/GADS/pull/145/commits/1ac38527d12822811cf856f95c3f1a73be0d0367))

provider/devices/common.go
gifflet: This was refactored to the following code:

```
		// Get the DeviceStreamSettings for the current device
		deviceStreamSettings, err := db.GetDeviceStreamSettings(dbDevice.UDID)
		if err != nil {
			// If there's an error (including not found), update the device with global settings
			err = updateDeviceWithGlobalSettings(dbDevice)
			if err != nil {
				logger.ProviderLogger.LogError("setupDevices", fmt.Sprintf("Failed to update device `%s` with global settings: %v", dbDevice.UDID, err))
				continue
			}
		} else {
			// Apply the retrieved stream settings
			dbDevice.StreamTargetFPS = deviceStreamSettings.StreamTargetFPS
			dbDevice.StreamJpegQuality = deviceStreamSettings.StreamJpegQuality
			dbDevice.StreamScalingFactor = deviceStreamSettings.StreamScalingFactor
		}
```

Does it seem ok? (full refactor available in this [commit](https://github.com/shamanec/GADS/pull/145/commits/1ac38527d12822811cf856f95c3f1a73be0d0367))

hub/gads-ui/src/components/Admin/GlobalSettings/GlobalSettings.js
gifflet: Of course. A snackbar offers a much better experience to the user. I will adjust that.

provider/devices/common.go
shamanec: Nice

hub/gads-ui/src/components/Admin/GlobalSettings/GlobalSettings.js
gifflet: Done ([commit](https://github.com/shamanec/GADS/pull/145/commits/f32160baa0d95c708e21c0c6d566171699a8c6c2))

provider/devices/common.go
gifflet: Really... this sleep is unnecessary inside the loop since we're already waiting 1 second using the ticker. The fix is available in the [commit](https://github.com/shamanec/GADS/pull/145/commits/039fc1a554e9e13d4761a87cffdbac16725f4e86)

shamanec: yeah, we're both stoopid :D

provider/devices/common.go
gifflet: > Even if you update the global stream settings, the actual update on the currently provisioned device happens only with the update-stream-settings API call.

I think the idea behind calling the API `update-stream-settings` is that the user can define specific stream settings for the device in use. The fact that updating global settings doesn't have the same behavior isn't necessarily a bad thing, considering that only sessions created after the update receive the current global stream settings. Am I correct?

>  they will not be reflected on a device that is already live - it will have to be re-provisioned - and if it is re-provisioned it will take the new global stream settings into account either way.

What are all the available actions that trigger device re-provisioning? I think the only way to trigger the device provisioning without the new goroutine   is by restarting the provider. By doing this, we can assume that the `setupDevices` function from `provider/devices/common.go` is executed, since that function doesn't run indefinitely in a goroutine like the others. 

When the device is already provisioned, and in live state, changing the global stream settings (assumming no device specific stream settings are available)  only takes affect when the goroutine runs. Otherwise, the provider must be restarted for the `setupDevices` function from `provider/devices/common.go` to run and so apply the current global settings to each device.

Does It all make sense to you?

shamanec: You could re-provision the device through the Admin panel or a manual disconnect/reconnect(sub-optimal 😅). What is that the current stream settings in WebDriverAgent and the Android stream apk get updated only in two scenarios: 1. When a device is re-provisioned 2. When you call the update-stream-settings endpoint. Right now even if this goroutine updates the stream settings in-memory in the map, these settings will not be applied on WebDriverAgent and the Android stream apk that are currently running. So lets say now you update the global stream settings from 15 fps to 30 fps, this goroutine will pick up the change, update the stream settings for the device in the map, but the actual stream will keep running at 15fps until the devices are re-provisioned. We either have to re-implement part of the logic from update-stream-settings here that will force an update to the streaming apps in this goroutine if theres a difference or maybe not use it at all. I admit my architecture is not very good 😂

gifflet: Initially, I had implemented the update of stream settings in WebDriverAgent and GADS Stream APK within the goroutine, but I ran into trouble updating these values in the UI, considering that the device was in use by a user.

So, if you think this update is really necessary inside the goroutine, we can work together to make it function as expected. What do you think? I can reintroduce the update of the stream settings in the goroutine, and after that, you can help me find the best approach to handle the update in the UI.

shamanec: Yeah, we could do that but if we also want it to update in the UI as well while a person is using it but that will require changes to the way device data is updated while in use, because right now no live updates are pushed, only on initial loading it gets data. I mean the stream quality will change but the UI wont reflect the changes. Maybe we can leave this for another PR and remove the goroutine from this one?

shamanec: In general I want to move away from MJPEG but implementing something like WebRTC is such a complex endeavour

gifflet: I think it's a good idea to work on UI live updates of the stream settings together with the goroutine implementation in another PR, but I couldn’t see the device stream settings update even after re-provisioning is triggered through the admin panel (without the goroutine). The scenario is:

```feature
Given I update the stream global settings
And I start the device re-provision from admin panel
When  I use a device from devices page
Then the current stream settings is present in the UI
```

Do you think this is an issue? If so, what can we do to fix this?

shamanec: Yeah, actually you are correct, I didn't analyze the changes properly. Currently you are doing the stream settings update in the `setupDevices` function which is called only once. I assume that if you move the `GetDeviceStreamSettings` logic to the `setupIOSDevice` and `setupAndroidDevice` functions it should be fine. You can probably make a function that accepts the db device pointer similar to [this](https://github.com/shamanec/GADS/blob/3db8d6d9e2ba976048758b611d39c74c80c85b4d/provider/devices/common.go#L824) one and do the stream settings updated in the functions I mentioned above when a device is being provisioned/re-provisioned

gifflet: I’ll move the device stream settings definition to setupIOSDevice and setupAndroidDevice, as you suggested. Since these functions run indefinitely in a goroutine, we should see the device stream settings getting updated after re-provisioning.

shamanec: Yeah, I assume as much but we'll see when you test it ;D

shamanec: lgtm, let me know if for some reason you can't merge it, but you should be able to

shamanec (approved):

gifflet:
@shamanec When a PR has many commits, I like to squash them into one. Would you agree with that?

shamanec:
6 is not that many but I am fine with it, I also squash sometimes 🤌

gifflet:
Thinking more carefully, I will proceed with the merge using a merge commit strategy because there are some snippets (the deleted goroutine) that we might need in the future as a reference for new work.

PR #158, opened 2025-02-20
Enable GADS Stream Settings Update via WebSocket

gifflet:
This PR enhances the Android device setup process by introducing a WebSocket-based mechanism to update GADS stream settings dynamically. Key changes include:

- WebSocket Integration for Stream Settings:
  - Implemented updateGadsStreamSettings(device), which sends updated stream settings via WebSocket.
  - Constructs a message containing targetFPS, jpegQuality, and scalingFactor and transmits it to the device.

- Android Device Setup Enhancement:
  - Calls updateGadsStreamSettings(device) after the device is connected to ensure correct streaming parameters.
  - Resets the device if updating stream settings fails.

### DIFF
--- a/provider/devices/common.go
+++ b/provider/devices/common.go
@@ -34,10 +34,6 @@ import (
 
 	"GADS/common"
 
-	"net/url"
-
-	"github.com/gobwas/ws"
-	"github.com/gobwas/ws/wsutil"
 	"go.mongodb.org/mongo-driver/bson"
 	"go.mongodb.org/mongo-driver/mongo"
 )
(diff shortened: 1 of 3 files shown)


provider/devices/common.go
```diff
@@ -995,3 +1006,25 @@ func applyDeviceStreamSettings(device *models.Device) error {
 
 	return nil
 }
+
+func updateGadsStreamSettings(device *models.Device) error {
```
shamanec, line 1010: Can you move this function to `android.go` and export it so it can be reused by the [UpdateDeviceStreamSettings](https://github.com/shamanec/GADS/blob/99315b781c5ed44bfae90ca7f3f8cb790dfc3a24/provider/router/routes.go#L349C6-L349C32) endpoint?

shamanec: In general this is stupid but it was the fastest way to implement stream quality updates along with WebDriverAgent. Probably a normal http server should be added to GADS-Stream so we can just fire a request without multiple hoops for the stupid websocket 😅

gifflet: Done

shamanec (approved):

PR #172, opened 2025-03-06
Enhance Device Setup and Reset Logic with Improved Concurrency Handling

### LINKED ISSUE #171: Multiple Provisioning Executions Cause Port Allocation Issues
### Description

When a disconnected device is reconnected, the associated Appium process is terminated due to the function that stops all Appium processes linked to the device. However, because the original Appium process was still active (as stdout was being read in a goroutine), this termination is abrupt and triggers the `resetLocalDevice` function.

This results in two different provisioning executions for the same device. The second execution cannot complete properly, potentially leading to multiple port allocations.

### Steps to Reproduce

1. Disconnect a device while an Appium session is active.
2. Ensure that the Appium process is still running because its stdout is being read.
3. Reconnect the device.
4. Observe that the function terminates all Appium processes linked to the device.
5. Notice that this abrupt termination triggers resetLocalDevice.
6. Check if two provisioning executions occur for the same device.
7. Verify that multiple ports may be allocated as a result.

### Expected Behavior

- The provisioning process should handle reconnections gracefully.
- Only a single provisioning execution should occur per device.
- Port allocation should be managed correctly without duplication.

### Actual Behavior

- Two provisioning executions occur for the same device.
- The second execution fails to complete properly.
- Multiple ports may be allocated incorrectly.

gifflet:
This PR improves the device setup and reset process by adding better concurrency handling and ensuring thread safety. Key changes include:

- Introduced SetupMutex for Devices:
  - Added SetupMutex to the Device struct to synchronize setup operations, preventing race conditions.
  - Applied SetupMutex.Lock() and SetupMutex.Unlock() in setupAndroidDevice and setupIOSDevice.

- Refactored resetLocalDevice Function:
  - Renamed to ResetLocalDevice and updated all references.
  - Improved logging by providing more descriptive reset reasons.

fixes #171

### DIFF
--- a/common/models/models.go
+++ b/common/models/models.go
@@ -79,6 +79,7 @@ type Device struct {
 	Logger           CustomLogger       `json:"-" bson:"-"` // CustomLogger object for the device
 	AppiumLogger     AppiumLogger       `json:"-" bson:"-"` // AppiumLogger object for logging appium actions
 	Mutex            sync.Mutex         `json:"-" bson:"-"` // Mutex to lock resources - especially on device reset
+	SetupMutex       sync.Mutex         `json:"-" bson:"-"` // Mutex for synchronizing device setup operations
 	GoIOSTunnel      tunnel.Tunnel      `json:"-" bson:"-"` // Tunnel obj for go-ios handling of iOS 17.4+
 	SemVer           *semver.Version    `json:"-" bson:"-"` // Semantic version of device for checks around the provider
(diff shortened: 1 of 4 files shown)


shamanec (approved):

shamanec:
Not sure we need stderr at all for Appium but we I'm fine with checking if the current changes will make setup more stable :)

PR #174, opened 2025-03-18
Implement User and Device Management for Workspaces

gifflet:
This PR introduces a workspace-based access control system, ensuring that users and devices are managed within specific workspaces.


hub/gads-ui/src/components/Admin/Users/UsersAdministration.js
shamanec, line 1: Users can be part of multiple workspaces but devices can be only assigned to one? Shouldn't it be the other way around? Or have both devices and users be able to be part of multiple workspaces?

hub/gads-ui/src/components/Admin/Workspaces/WorkspacesAdministration.js
```diff
@@ -0,0 +1,290 @@
...
+                                </Button>
+                                <Button variant='outlined' onClick={handleCloseModal}>
+                                    Cancel
+                                </Button>
+                            </Stack>
+                        </Stack>
+                    </Box>
+                </Modal>
+
+                <Modal open={openDeleteModal} onClose={handleCloseDeleteModal}>
```
shamanec, line 247: We already have a reusable dialog for such actions applied across the whole app - you can check how it is used [here](https://github.com/shamanec/GADS/blob/c91ff7c85346b3bbb56da223bdaeab0096bf0308/hub/gads-ui/src/components/Admin/Users/UsersAdministration.js#L220)

hub/router/routes.go
```diff
@@ -224,7 +216,7 @@ func UpdateUser(c *gin.Context) {
 		return
 	}
 
-	if user == (models.User{}) {
+	if user.Username == "" || user.Password == "" || (user.Role == "user" && len(user.WorkspaceIDs) == 0) {
```
shamanec, line 219: We want to update users without mandatorily supplying a password in the UI, right now if you try to update a user and you do not add a password value, update will fail

common/db/workspace.go
```diff
@@ -0,0 +1,200 @@
...
+
+import (
+	"GADS/common/models"
+	"context"
+
+	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/mongo/options"
+)
+
+func AddWorkspace(workspace *models.Workspace) error {
```
shamanec, line 11: If we are not going to keep those objects somewhere in memory that has to be updated we don't really to use a pointer, it doesn't really matter though as long as it works

common/models/models.go
```diff
@@ -141,3 +144,15 @@ type StreamSettings struct {
...
+
+type Workspace struct {
+	ID          string `json:"id" bson:"_id,omitempty"`
+	Name        string `json:"name" bson:"name"`
+	Description string `json:"description" bson:"description"`
+	IsDefault   bool   `json:"is_default" bson:"is_default"`
+}
+
+// GenerateUUID generates a new UUID for the workspace
+func (w *Workspace) GenerateUUID() {
```
shamanec, line 156: I am not really sure we need this, Mongo automatically creates arbitrary `_id` when a document is added

common/db/workspace.go
```diff
@@ -0,0 +1,200 @@
...
+	var workspaces []models.Workspace
+	collection := mongoClient.Database("gads").Collection("workspaces")
+
+	cursor, err := collection.Find(mongoClientCtx, bson.M{})
+	if err != nil {
+		return workspaces
+	}
+	defer cursor.Close(mongoClientCtx)
+
+	for cursor.Next(mongoClientCtx) {
```
shamanec, line 52:
```suggestion
	if err := cursor.All(mongoClientCtx, &workspaces); err != nil {
	
	}
```
cleaner option, we want data in the DB to be stable enough to confidently decode all workspaces into the slice directly instead returning only ones that can be "decoded"

common/db/workspace.go
```diff
@@ -0,0 +1,200 @@
...
+	count, err := collection.CountDocuments(mongoClientCtx, filter)
+	if err != nil {
+		return false
+	}
+	return count > 0
+}
+
+func WorkspaceHasUsers(id string) bool {
+	collection := mongoClient.Database("gads").Collection("users")
+	filter := bson.M{"workspace_ids": id}
```
shamanec, line 75: Will this filter work properly if we have multiple workspaces since its an array? Not really sure, just asking :D

common/db/workspace.go
```diff
@@ -0,0 +1,200 @@
...
+	filter := bson.M{"is_default": true}
+
+	err := collection.FindOne(context.TODO(), filter).Decode(&workspace)
+	if err != nil {
+		return models.Workspace{}, err
+	}
+	return workspace, nil
+}
+
+func GetWorkspacesPaginated(page, limit int, search string) ([]models.Workspace, int64) {
```
shamanec, line 119: I highly doubt someone will scale GADS so much that they have pageS of workspaces but we can have it just in case 😅

common/models/models.go
```diff
@@ -141,3 +144,15 @@ type StreamSettings struct {
 	ScalingFactorAndroid int `json:"scaling_factor_android,omitempty" bson:"scaling_factor_android"`
 	ScalingFactoriOS     int `json:"scaling_factor_ios,omitempty" bson:"scaling_factor_ios"`
 }
+
+type Workspace struct {
+	ID          string `json:"id" bson:"_id,omitempty"`
```
shamanec, line 149: I am 99% sure that the _id cannot be empty unless explicitly set to empty in some way, I guess `omitempty` is redundant

hub/auth/auth.go
shamanec, line 1: Would you mind removing the commented code from this file?

hub/gads-ui/src/components/Admin/Users/UsersAdministration.js
```diff
@@ -136,6 +184,41 @@ function NewUser({ handleGetUserData }) {
                             <MenuItem value='admin'>Admin</MenuItem>
                         </TextField>
                     </FormControl>
+                    <FormControl fullWidth required>
```
shamanec, line 187: I wonder if something like [this](https://mui.com/material-ui/react-select/#checkmarks) wouldn't be more appropriate? What do you think?

hub/router/routes.go
```diff
@@ -171,7 +171,7 @@ func AddUser(c *gin.Context) {
...
```
shamanec, line 174: Same as [here](https://github.com/shamanec/GADS/pull/174/files#r2005383127)

provider/provider.go
```diff
@@ -59,6 +61,23 @@ func StartProvider(flags *pflag.FlagSet) {
 	logger.SetupLogging(logLevel)
 	logger.ProviderLogger.LogInfo("provider_setup", fmt.Sprintf("Starting provider on port `%v`", config.ProviderConfig.Port))
 
+	// Check if the default workspace exists
+	defaultWorkspace, err := db.GetDefaultWorkspace()
```
shamanec, line 65: I don't think there is a point to do this for the provider, you can't even start a provider instance without starting a hub instance to create the config :P

shamanec (changes requested):
some nitpicks here and there, great job!

hub/gads-ui/src/components/Admin/Workspaces/WorkspacesAdministration.js
```diff
@@ -0,0 +1,290 @@
...
+        setCurrentWorkspace(null);
+        setNewWorkspace({ name: '', description: '' });
+    };
+
+    const handleUpdateWorkspace = async () => {
+        try {
+            await api.put(`/admin/workspaces`, {
+                id: currentWorkspace.id,
+                name: currentWorkspace.name,
+                description: currentWorkspace.description,
```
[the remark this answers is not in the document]
gifflet: The fix has been applied to the backend

hub/gads-ui/src/components/Admin/Users/UsersAdministration.js
shamanec: Nah, the current implementation is fine by me, was just wondering. It can always be changed if the need arises :)

hub/gads-ui/src/components/Admin/Workspaces/WorkspacesAdministration.js
gifflet: Done in [commit](https://github.com/shamanec/GADS/pull/174/commits/f442d635999eae72c3da43fc681d2be449b18be9)

hub/router/routes.go
gifflet: It does make a lot of sense. I will adjust it...

gifflet: Done in [commit](https://github.com/shamanec/GADS/pull/174/commits/60171a9866d44d9e4171d88eae61717a4ad86d1c)

common/models/models.go
```diff
@@ -34,7 +34,7 @@ func (a ByUDID) Less(i, j int) bool { return a[i].UDID < a[j].UDID }
 
 type User struct {
 	Username     string   `json:"username" bson:"username"`
-	Password     string   `json:"password" bson:"password"`
+	Password     string   `json:"password" bson:"password,omitempty"`
```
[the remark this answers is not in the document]
gifflet: During a user update, if the password is empty, it gets erased unless `omitempty` is set.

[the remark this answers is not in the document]
gifflet: That was the reason I implemented the password validation in user update before.

common/models/models.go
gifflet: Changes in [commit](https://github.com/gifflet/GADS/commit/68c3e94a31bc8cb37c9de49884f9fbde25605bdb)

common/db/workspace.go
gifflet: Changes in [commit](https://github.com/gifflet/GADS/commit/0e26986312a67d7b649462fd5c13edff7d2de3bd)

common/db/workspace.go
gifflet: Yes, when we query an array field with a single value, MongoDB will match any document where the array contains that value ([reference](https://www.mongodb.com/docs/manual/tutorial/query-arrays/#query-an-array-for-an-element))

common/db/workspace.go
gifflet: In this [commit](https://github.com/gifflet/GADS/commit/68c3e94a31bc8cb37c9de49884f9fbde25605bdb), we are using a pointer to the workspace because we apply the newly generated workspace ID from MongoDB. I observed that after creating a workspace and trying to use it to associate users or workspaces, its ID value is an empty string ("") since MongoDB doesn’t assign an ID as fast as we try to use its value.

```go
func AddWorkspace(workspace *models.Workspace) error {
 	collection := mongoClient.Database("gads").Collection("workspaces")
 	result, err := collection.InsertOne(mongoClientCtx, workspace)
 	if err != nil {
 		return err
 	}
 	workspace.ID = result.InsertedID.(primitive.ObjectID).Hex()
 	return nil
 }
```

hub/auth/auth.go
gifflet: Done. ([commit](https://github.com/gifflet/GADS/commit/285ba89c614916d7ec71b6e7159b83f4162243e5))

hub/gads-ui/src/components/Admin/Users/UsersAdministration.js
gifflet: Yeah, you are right! The user experience will be better, and the UI will fit better with the Select element as more workspaces are selected. Done in this [commit](https://github.com/gifflet/GADS/commit/11b90bb1b1a48c27197ab595fb27b23ac27149ce).

hub/router/routes.go
gifflet: I'm not sure what should be changed in this code section... This code is part of user creation, so it's expected that username, password, and possible workspaces are valid, right?

provider/provider.go
gifflet: In a scenario where the user starts the provider before the hub and the default workspace doesn't exist, the provider needs to create the default workspace. Additionally, the provider should assign its devices to the default workspace if they have no existing assignment.

common/db/workspace.go
shamanec: Weird, the MongoDB driver [docs](https://www.mongodb.com/docs/manual/reference/method/db.collection.insertOne/) claim it should be returned... Maybe we should update the package hoping it won't break stuff :D

hub/router/routes.go
shamanec: Ah, you are right, I didn't look properly :D

provider/provider.go
shamanec: Yeah, that makes sense but you are required to have a provider configuration added to the hub to be able to start it, if you don't provide `--nickname` flag or its invalid, then the provider instance will not boot up at all? :)

gifflet: Right, but I’m thinking of a scenario where the user is running GADS from previous versions and already has MongoDB configured with all the data required for GADS to run

gifflet: If we agree that only the hub should create the default workspace, as we can have many providers and potential issues in the future regarding workspace creation on the provider side, I can remove this feature from the provider.

shamanec: You are right, it is better to tighten our pants with this. At some point after multiple releases when this should no longer be a case, we might remove it but let's leave it for now 👌

common/models/models.go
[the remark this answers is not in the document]
shamanec: This was handled in the UpdateUser function by checking if an empty password string is coming from the frontend, but if this works as you say - lets have it like this :)

common/db/workspace.go
gifflet: The docs state that the ID value is returned through the `InsertedID` field, so the current implementation is correct, right?

common/models/models.go
gifflet: When not using `omitempty`, the value stored in the database is `""` because we are not setting its value, and the serialization is not removing the empty string.

shamanec: Yes, because we are using `string` which is not the idiomatic type for MongoDB documents, it should be `primitive.ObjectID`. Unfortunately I've written some code where I didn't read the docs comprehensively or spent much time thinking on it and we already have some string types for `_id` here and there 😓 🤣

common/db/workspace.go
shamanec: yes, I believe so, I can see that the ID value is returned in the result currently 👌

gifflet:
> You should apply the same background and button colors as across the other UI

Updated in [commit](https://github.com/shamanec/GADS/pull/174/commits/4b16e42837ad5cb523924bd7ebf444c8db953034).

gifflet:
Docs updated in [commit](https://github.com/shamanec/GADS/pull/174/commits/224c1683fe647955842cd1416cee59ae68c1aa83).

shamanec (approved):

PR #183, opened 2025-04-22
Support only custom WebDriverAgent fork

shamanec:
* Remove support for mainstream WebDriverAgent for device remote control
* Remove provider configuration property related to custom WebDriverAgent
* Update readme to include info only about custom WebDriverAgent fork
* Remove button for refreshing Appium session in remote control for iOS devices

### DIFF
--- a/docs/provider.md
+++ b/docs/provider.md
@@ -132,7 +132,10 @@ Example installation command for Ubuntu -  `sudo apt install usbmuxd`.
 
 #### WebDriverAgent ipa
 You need to prepare and upload a signed `WebDriverAgent` ipa file from the hub UI in `Admin > Files`  
-GADS supports only the custom WebDriverAgent from my fork.  
+GADS supports only WebDriverAgent from my [fork](https://github.com/shamanec/WebDriverAgent).  
+The fork has optimizations for the mjpeg video stream and additional endpoints for faster tap/swipe interactions that are not available in the mainstream repo.  
+Additionally those endpoints require different coordinates for interaction from mainstream WDA which forces separate handling for the remote control which is too much work.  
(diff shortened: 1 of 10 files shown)


docs/provider.md
```diff
@@ -132,8 +132,7 @@ Example installation command for Ubuntu -  `sudo apt install usbmuxd`.
 
 #### WebDriverAgent ipa
 You need to prepare and upload a signed `WebDriverAgent` ipa file from the hub UI in `Admin > Files`  
-By using the custom WebDriverAgent from my fork you can have faster tap/swipe interactions on iOS devices (the provider configuration should be set to use the custom WebDriverAgent through hub UI).  
-You can use mainstream WebDriverAgent as well  
+GADS supports only the custom WebDriverAgent from my fork.
```
gifflet, line 135: It might be beneficial to clarify the rationale behind dropping support for the mainstream WebDriverAgent. This could prevent potential confusion for contributors or users familiar with previous versions. A brief note explaining why only the custom fork is supported (e.g., technical limitations, compatibility, performance reasons) would enhance documentation transparency.

gifflet (changes requested):

docs/provider.md
shamanec: Not really sure the need to explain, if someone has issues with this can always consider another project but I added more info [here](https://github.com/shamanec/GADS/pull/183/commits/4be4a03ea7d436639eb222e7199c74b6ff196c21) 😅 If it becomes a problem we could introduce one WDA for remote control and one for Appium tests, I believe two can run at the same time on a device with different ports

gifflet (approved):

PR #195, opened 2025-05-19
Add Support for Origin-Based JWT Secret Keys and Key Management APIs

gifflet:
This PR introduces a flexible and scalable JWT authentication mechanism based on request origin, allowing the use of different secret keys per origin. This enhancement enables support for multi-tenant and multi-client applications while ensuring secure and maintainable JWT management.

🛠️ Key Features:

**- Origin-Based JWT Secret Support**
JWT tokens are now signed and validated using secrets scoped by the request’s origin, supporting isolation and security per client/system.

**- MongoDB-Backed Secret Store**
Persistent storage for secret keys with metadata (origin, status, timestamps, etc.).

**- In-Memory Cache**
For high-performance lookups and minimal DB hits during token validation and generation.

**- JWT Enhancements**
JWT tokens can now be:
    - Encrypted and decrypted using the appropriate origin-based key
    - Validated with fallback logic in case of legacy/default keys


hub/auth/secretcache.go
```diff
@@ -0,0 +1,173 @@
...
+	// If there are no keys at all, create a default one
+	if len(secretKeys) == 0 {
+		// Generate a random key for security
+		randomKey, err := generateRandomKey(32) // 32 bytes for good security
+		if err != nil {
+			return err
+		}
+
+		// Create a new default key
+		defaultKey := &SecretKey{
```
shamanec, line 67: We are not supplying `user_identifier_claim` here for the default key, but when editing it via the hub the User Identifier Claim is mandatory and has to be provided - should that be expected? Maybe some default claim should be applied here as well?

hub/router/secrets.go
```diff
@@ -0,0 +1,350 @@
...
+			"origin":                  key.Origin,
+			"is_default":              key.IsDefault,
+			"created_at":              key.CreatedAt,
+			"updated_at":              key.UpdatedAt,
+			"user_identifier_claim":   key.UserIdentifierClaim,
+			"tenant_identifier_claim": key.TenantIdentifierClaim,
+		})
+	}
+
+	c.JSON(http.StatusOK, gin.H{"secret_keys": response})
```
shamanec, line 39: won't this work as well while being less verbose?
`
	for i := range keys {
		keys[i].Key = ""
	}

	c.JSON(http.StatusOK, gin.H{"secret_keys": keys})
`

hub/router/secrets.go
```diff
@@ -0,0 +1,350 @@
...
+func AddSecretKey(c *gin.Context) {
+	// Get username from user claims for audit
+	username, exists := getUsernameFromContext(c)
+	if !exists {
+		c.JSON(http.StatusInternalServerError, gin.H{"error": "Failed to get user information"})
+		return
+	}
+
+	// Parse request body
+	var request struct {
```
shamanec, line 52: Can't we reuse the `SecretKey` struct here and update the necessary fields after before adding it in the DB?

hub/auth/secretcache.go
gifflet: Fixed in [commit](https://github.com/shamanec/GADS/pull/195/commits/0fb01fcb89f9ee4e0d808776e6449fbd85e646c8)

hub/router/secrets.go
```diff
@@ -0,0 +1,350 @@
...
+		UserIdentifierClaim:   request.UserIdentifierClaim,
+		TenantIdentifierClaim: request.TenantIdentifierClaim,
+	}
+
+	// Add secret key to database
+	store := auth.NewSecretStore(db.GlobalMongoStore.GetDefaultDatabase())
+	err := store.AddSecretKey(secretKey, username, request.Justification)
+	if err != nil {
+		if err == auth.ErrDuplicateOrigin {
+			c.JSON(http.StatusBadRequest, gin.H{"error": "An origin with this name already exists"})
```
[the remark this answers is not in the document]
gifflet: During the first GADS startup, a default secret key is created and set as the default. This way, the user doesn’t need to deal with secret key registration, assuming they plan to use a single secret key for all incoming hub requests. If a request comes from an unregistered origin, the default secret key is used to validate its bearer token (if present)

hub/router/secrets.go
gifflet: Yeah, it worked but the secret key that is always blank is unnecessary added:

payload returned:

```json
{
    "secret_keys": [
        {
            "id": "682b3571b8795398c5716b95",
            "origin": "default",
            "key": "",
            "is_default": false,
            "created_at": "2025-05-19T13:43:13.254Z",
            "updated_at": "2025-05-19T13:44:21.729Z",
            "disabled": false
        },
        {
            "id": "682b35b5b8795398c5716b97",
            "origin": "http://192.168.3.80:10000",
            "key": "",
            "is_default": true,
            "created_at": "2025-05-19T13:44:21.731Z",
            "updated_at": "2025-05-19T13:44:21.731Z",
            "disabled": false,
            "user_identifier_claim": "username"
        }
    ]
}
```

verbose implementation result:

```json
{
    "secret_keys": [
        {
            "created_at": "2025-05-19T13:43:13.254Z",
            "id": "682b3571b8795398c5716b95",
            "is_default": false,
            "origin": "default",
            "tenant_identifier_claim": "",
            "updated_at": "2025-05-19T13:44:21.729Z",
            "user_identifier_claim": ""
        },
        {
            "created_at": "2025-05-19T13:44:21.731Z",
            "id": "682b35b5b8795398c5716b97",
            "is_default": true,
            "origin": "http://192.168.3.80:10000",
            "tenant_identifier_claim": "",
            "updated_at": "2025-05-19T13:44:21.731Z",
            "user_identifier_claim": "username"
        }
    ]
}
```

hub/router/secrets.go
gifflet: It’s not recommended to use a database model to parse the request body, as the user might modify fields that shouldn’t be changed, like the ID.

hub/router/secrets.go
shamanec: I don't think we need to save data from those requests in particular but okay, lets leave it as is :D

shamanec:
This is too much code that will take me days to review properly but I did look over it and I put my trust in you :D GJ 🚀

shamanec (approved):

shamanec (approved):

PR #206, opened 2025-06-01
Implement Android device remote control without Appium

shamanec:
* Add new single Android APK containing remote control server code, IME, mjpeg and webrtc streaming
* Rework Android device setup to use the new apk for IME and streaming
* Rework the control functions to use the Android remote control server instead of Appium
* Left Appium endpoints related code in case fallback has to be implemented though in theory it shouldn't be needed


provider/devices/android.go
```diff
@@ -166,11 +200,12 @@ func addGadsStreamRecordingPermissions(device *models.Device) error {
...
 }
 
-// Start the gads-stream app using adb
-func startGadsStreamApp(device *models.Device) error {
+// Start the GADS video streaming service using adb
+func startGadsAndroidStreaming(device *models.Device) error {
 	logger.ProviderLogger.LogInfo("android_device_setup", fmt.Sprintf("Starting GADS-stream app on `%s`", device.UDID))
 
 	cmd := exec.CommandContext(device.Context, "adb", "-s", device.UDID, "shell", "am", "start", "-n", GetStreamServiceActivityName(device))
+	fmt.Printf("Starting activity with `%v`", cmd.Args)
```
gifflet: replace with debug log

provider/devices/android.go
```diff
@@ -341,21 +378,21 @@ func UpdateGadsStreamSettings(device *models.Device) error {
...
 func GetStreamServiceName(device *models.Device) string {
 	if device.UseWebRTCVideo {
-		return "com.gads.webrtc/.ScreenCaptureService"
+		return "com.gads.settings/.WebRTCScreenCaptureService"
 	}
-	return "com.shamanec.stream/.ScreenCaptureService"
+	return "com.gads.settings/.ScreenCaptureService"
 }
 
 func GetStreamServicePackageName(device *models.Device) string {
```
gifflet, line 386:
```suggestion
func GetStreamServicePackageName(device *models.Device) string {
	return "com.gads.settings"
}
```

gifflet (changes requested):

provider/devices/android.go
shamanec: oops :D fixed [here](https://github.com/shamanec/GADS/pull/206/commits/acd4599c28699bc9b74ee161c29768c64d76b5d3)

provider/devices/android.go
shamanec: done [here](https://github.com/shamanec/GADS/pull/206/commits/acd4599c28699bc9b74ee161c29768c64d76b5d3)

provider/devices/common.go
```diff
@@ -304,83 +313,65 @@ func setupAndroidDevice(device *models.Device) {
...
+		err = UninstallApp(device, "com.gads.settings")
 		if err != nil {
-			logger.ProviderLogger.LogError("android_device_setup", fmt.Sprintf("Could not uninstall GADS mjpeg stream app from Android device - %v:\n %v", device.UDID, err))
-			ResetLocalDevice(device, "Failed to uninstall GADS mjpeg stream app from Android device.")
+			logger.ProviderLogger.LogError("android_device_setup", fmt.Sprintf("Could not uninstall GADS Settings app from Android device - %v:\n %v", device.UDID, err))
+			ResetLocalDevice(device, "Failed to uninstall GADS Settings from Android device.")
 		}
 		time.Sleep(3 * time.Second)
 	}
+	logger.ProviderLogger.LogDebug("android_device_setup", fmt.Sprintf("Checked for and uninstalled existing GADS Android apps on device `%v` if they were present", device.UDID))
```
gifflet: I just checked and noticed that if GADS Stream (which is running) isn’t uninstalled, GADS Settings ends up failing to access the device screen. So I believe that, for this new app to work properly, it might be better to uninstall all the older GADS apps as well. What do you think?

shamanec: Great catch! I completely forgot people might have leftover running apps from previous releases 😅 fixed [here](https://github.com/shamanec/GADS/pull/206/commits/ae04820a11c3dc1b49247945a6a783562674c0d0)

gifflet: We’re missing com.gads.gads_ime, right? The keyboard is also included in GADS Settings, isn’t it?

shamanec: It doesn't really matter because the correct one will be enabled when provisioning, but I will add it for uninstalling as well 👌

shamanec: Done [here](https://github.com/shamanec/GADS/pull/206/commits/dc36690a95e7bfbe9c9ac11674ffd58c12093ae6)

PR #227, opened 2025-07-24
Introduce Appium plugin for structured logging

shamanec:
* Update provider setup to check for and install the GADS Appium plugin on the host
* Introduce new provider endpoints to accept logs and other relevant data from the GADS Appium plugin
* Remove the old Appium logger implementation
* Cleared up the Appium setup functions since we no longer need to parse the terminal output
* Removed other obsolete code related to old Appium logging implementation


common/db/log.go
```diff
@@ -39,11 +42,17 @@ func (m *MongoStore) GetProviderLogs(collectionName string, logLimit int) ([]mod
 	return GetDocuments[models.ProviderLog](m.Ctx, coll, bson.D{{}}, findOptions)
 }
 
-func (m *MongoStore) GetAppiumSessionLogs(collectionName, sessionID string) ([]models.AppiumLog, error) {
+func (m *MongoStore) GetAppiumSessionLogs(collectionName, sessionID string) ([]models.AppiumPluginLog, error) {
 	coll := m.GetCollectionWithDB(appiumLogDB, collectionName)
 	findOptions := options.Find()
 	findOptions.SetSort(bson.D{{Key: "ts", Value: -1}})
```
gifflet: It should be `timestamp`, right?

provider/devices/common.go
```diff
@@ -169,20 +169,17 @@ func setupDevices() {
 			appiumCollectionIndexModel := mongo.IndexModel{
 				Keys: bson.D{
 					{
-						Key: "ts", Value: constants.SortAscending},
+						Key: "ts", Value: constants.SortAscending,
```
gifflet: `timestamp`?

provider/provider.go
```diff
@@ -87,6 +87,29 @@ func StartProvider(flags *pflag.FlagSet, resourceFiles embed.FS) {
...
+		if !providerutil.CheckAppiumPluginInstalledNPM() {
+			logger.ProviderLogger.LogInfo("provider_setup", "Installing GADS Appium plugin globally on host NPM")
+			err = providerutil.InstallAppiumPluginNPM()
+			if err != nil {
+				log.Fatalf("Failed to install GADS Appium plugin on NPM - %s", err)
+			}
+			logger.ProviderLogger.LogInfo("provider_setup", "Successfully installed GADS Appium plugin globally on host NPM")
+		} else {
+			logger.ProviderLogger.LogInfo("provider_setup", "GADS Appium plugin already installed on host NPM")
+		}
```
gifflet, line 101: Is it really necessary to install the plugin globally, since it’s only needed by Appium?

provider/providerutil/providerutil.go
```diff
@@ -112,6 +113,74 @@ func AppiumAvailable() bool {
...
+	for _, line := range lines {
+		if strings.Contains(line, "appium-gads@") {
+			parts := strings.Split(line, "@")
+			if len(parts) == 2 {
+				return strings.TrimSpace(parts[1]), nil
+			}
+		}
+	}
+	return "", fmt.Errorf("Could not get GADS Appium plugin version on NPM")
+}
```
gifflet, line 143: Should it be removed?

gifflet (changes requested):

provider/providerutil/providerutil.go
shamanec: I've added this intentionally, probably on next version each version will have hardcoded supported plugin version so the provider can update automatically if needed. We do not use it now but we will probably use it in the future :)

provider/provider.go
shamanec: Well maybe we can extrapolate where Appium is installed and then install the plugin only there, but I dont see a problem to have it globally installed 🤷

common/db/log.go
shamanec: We are not using this yet but good catch, fixed [here](https://github.com/shamanec/GADS/pull/227/commits/2bfcf85fb8848d04ccf71e566b91ca350d191812)

provider/devices/common.go
shamanec: Damn, fixed [here](https://github.com/shamanec/GADS/pull/227/commits/2bfcf85fb8848d04ccf71e566b91ca350d191812)

gifflet (approved):

PR #235, opened 2025-09-10
Add Appium session reports

shamanec:
This PR introduces:

* GADS Appium plugin code
* Provider handling of Appium plugin session action logs storage
* Hub handling of Appium build/session reports

When the Appium plugin creates sessions with `gads:buildId` capability provided it will send action logs to the provider for storage in MongoDB for each command executed. Logs are stored per tenant and are accessible for any user from the current tenant.
Hub aggregates the logs into suitable for presentation in the UI reports. The reporting is still a work in progress but it is already quite a lot of code hence this PR

### DIFF
--- a/common/db/log.go
+++ b/common/db/log.go
@@ -11,6 +11,7 @@ package db
 
 import (
 	"GADS/common/models"
+	"fmt"
 
 	"go.mongodb.org/mongo-driver/bson"
 	"go.mongodb.org/mongo-driver/mongo"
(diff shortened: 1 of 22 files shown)


common/db/log.go
```diff
@@ -42,17 +45,299 @@ func (m *MongoStore) GetProviderLogs(collectionName string, logLimit int) ([]mod
...
+		return nil, fmt.Errorf("failed to find session logs: %w", err)
+	}
+	defer cursor.Close(m.Ctx)
+
+	var sessionLogs []models.AppiumPluginSessionLog
+	if err = cursor.All(m.Ctx, &sessionLogs); err != nil {
+		return nil, fmt.Errorf("failed to decode session logs: %w", err)
+	}
+
+	return sessionLogs, nil
```
gifflet, line 342: I think it would be really valuable if we could filter the session logs by action. For example, a user might not want to see events from the `Other` action.

hub/router/reports.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	sessionLogs, err := db.GlobalMongoStore.GetSessionLogs(tenant, sessionID)
+	if err != nil {
+		api.GenericResponse(c, 500, err.Error(), nil)
+		return
+	}
+
+	api.GenericResponse(c, 200, "Got session logs", sessionLogs)
+}
```
gifflet, line 90: Regarding the common/db/log.go review: it would be great if the filter options (like Action) could be optional, with a default that filters out Other, since most users likely won’t be interested in that event type. Also, having the substring of Additional info as an optional filter would allow searching for a specific payload in an event, making it easier to understand issues during execution when errors occur or something doesn’t work as expected.

provider/providerutil/providerutil.go
```diff
@@ -170,8 +186,32 @@ func CheckAppiumPluginInstalled() bool {
...
+	if err != nil {
+		logger.ProviderLogger.LogError("provider_setup", fmt.Sprintf("Failed to uninstall GADS Appium plugin - %s", string(out)))
+		return err
+	}
+	return nil
+}
+
+// Update the GADS Appium plugin on Appium
+func UpdateAppiumPlugin() error {
+	cmd := exec.Command("appium", "plugin", "update", "gads")
```
gifflet: Running this function might install a plugin version incompatible with a specific GADS installation. Does that make sense? Should plugin installation be handled only by the install and uninstall functions?

provider/router/appium_plugin.go
```diff
@@ -78,6 +101,7 @@ func AppiumPluginRemoveSession(c *gin.Context) {
 	api.GenericResponse(c, http.StatusNotFound, fmt.Sprintf("Device with udid `%s` not found", udid), nil)
 }
 
+
```
gifflet: Unnecessary empty line

common/db/log.go
shamanec: We could but I think I've handled most of the actual commands in the plugin code, `Other` is usually `screenshot`, `timeouts` and `getSessions` that are called by the Appium inspector most of the time, so `Other` should be a rare case when actually running tests. In general logs will be stored only when `gads:buildId` is provided which is not expected to be done through the inspector. I could add some filtering options though on a later PR - reports are still in progress, hence the `develop` branch 😅

hub/router/reports.go
shamanec: I'll think about this in a consecutive PR, maybe filtering can be done on UI side, even a long single E2E test session should not have thaaaat many commands executed to make UI filtering heavy, although I noticed that `wdio` has some internal logic to call automatically `findElements` if `findElement` fails which could lead to a bigger session log result

provider/providerutil/providerutil.go
shamanec: Actually this function is obsolete and I haven't removed it, it is not used anywhere. You are correct though - currently when we update the plugin we do uninstall and then install exactly for this reason! 🚀

provider/router/appium_plugin.go
shamanec: Removed [here](https://github.com/shamanec/GADS/pull/235/commits/fe5f8a537102a22ad6cb2fbf0cbfde7f4d2c535f)

provider/providerutil/providerutil.go
shamanec: Removed [here](https://github.com/shamanec/GADS/pull/235/commits/fe5f8a537102a22ad6cb2fbf0cbfde7f4d2c535f)

PR #237, opened 2025-09-26
Resolve pagination bug when filtering by tenant

gifflet:
The GetWorkspaces() endpoint had a pagination bug when filtering by tenant:
  - Tenant filtering was happening AFTER pagination instead of before
  - This caused missing results when tenant workspaces were distributed across multiple pages
  - Example: 6 total workspaces, 3 for specific tenant, with 5 items per page
    - One tenant workspace on page 2 would be missing when requesting page 1 with tenant filter
    - Only showed 2 workspaces instead of all 3 for that tenant

  ## Solution
  - Move tenant filtering to database level (MongoDB aggregation pipeline)
  - Add `tenant` parameter to `GetWorkspacesWithDeviceCount()` function
  - Remove post-query filtering logic from router
  - Ensure correct pagination and total count calculation

  ## Changes
  - **common/db/workspace.go**: Add tenant parameter and database-level filtering
  - **hub/router/workspace.go**: Pass tenant parameter to database function, remove post-query filtering

  ## Impact
  - Fixes incorrect pagination results when filtering by tenant
  - Improves query performance by filtering at database level
  - Ensures accurate total count for pagination

### DIFF
--- a/common/db/workspace.go
+++ b/common/db/workspace.go
@@ -150,7 +150,7 @@ func (m *MongoStore) GetUserWorkspaces(username string) []models.Workspace {
 	return userWorkspaces
 }
 
-func (m *MongoStore) GetWorkspacesWithDeviceCount(page, limit int, search string) ([]models.WorkspaceWithDeviceCount, int64, error) {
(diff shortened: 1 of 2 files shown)


shamanec:
🙌

PR #239, opened 2025-10-29
Add app installation support for Tizen and webOS platforms

gifflet:
This PR implements comprehensive app installation functionality for Tizen and webOS smart TV platforms, completing the platform support matrix for TV devices.

### Key Features

 **Tizen Support:**
- Certificate-based .wgt package signing and installation
- Automatic certificate detection from `~/SamsungCertificate/` directory
- Extraction and packaging of .wgt archives
- Integration with Tizen CLI tools (`tizen package` and `tizen install`)

 **webOS Support:**
- Direct .ipk installation support
- Source code packaging and installation workflow
- Integration with webOS CLI tools (`ares-package` and `ares-install`)

### DIFF
--- a/provider/devices/common.go
+++ b/provider/devices/common.go
@@ -10,7 +10,6 @@
 package devices
 
 import (
-	"archive/zip"
 	"bufio"
 	"bytes"
 	"context"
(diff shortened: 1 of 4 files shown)


provider/devices/common.go
```diff
@@ -1237,3 +1250,52 @@ func getConnectedDevicesTizen() []string {
 
 	return devices
 }
+
+func extractZipToDir(zipPath, destDir string) error {
```
shamanec: Maybe we should put this in the `common` package of the project? Also we already have a an unzipping function in `common/utils/archive.go`, can you check if it can be used? :)

gifflet: The UnzipInMemory function in common/utils/archive.go only extracts the first file if it’s not a directory. This means it extracts all files only when the zip contains a root folder. The extractZipToDir function, however, extracts all files regardless of whether there’s a root folder inside the zip. I’ll move extractZipToDir to common/utils/archive.go, ok?

gifflet: Done in [commit](https://github.com/shamanec/GADS/pull/239/commits/cea8ea0ee1cf4fc247d1c97391745df58d1141e1)

shamanec: I mean the `Unzip` not `UnzipInMemory` function but thats fine, gj

PR #258, opened 2026-02-26
feat(android): add ADB over TCP/IP support for remote debugging

mogilevich:
## Summary

Allow users to connect to Android devices directly via ADB over Wi-Fi while GADS manages the USB connection.

**Flow:** take device → enable ADB TCP/IP → get `adb connect IP:5555` command → work remotely → release device → provider auto-restores USB mode.

## Changes

- `POST /device/:udid/adb-tcpip/enable` — enables TCP/IP mode on port 5555, returns device IP and connect command
- `POST /device/:udid/adb-tcpip/disable` — restores USB-only mode
- Re-establishes ADB port forwards (stream, IME, remote server) after daemon restart caused by `adb tcpip` / `adb usb`
- `AdbTcpIpTransitioning` flag on device prevents false device resets while ADB daemon is intentionally restarting
- WebRTC H264 stream reconnects transparently without closing the browser peer connection (`reconnectStream` in `writeFrames`)
- Hub proxy: `enable` is gated — only works when device is actively in use by the requesting user
- Hub proxy: `disable` is gated — only the session owner can disable (or anyone when device is free)
- Hub: auto-disables ADB TCP/IP when the device control WebSocket session ends

### DIFF
--- a/common/models/models.go
+++ b/common/models/models.go
@@ -99,6 +99,7 @@ type Device struct {
 	SemVer                  *semver.Version    `json:"-" bson:"-"` // Semantic version of device for checks around the provider
 	InitialSetupDone        bool               `json:"-" bson:"-"` // On provider startup some data is prepared for devices like logger, Mongo collection, etc. This is true if all is done
 	DeviceAddress           string             `json:"-" bson:"-"`
+	AdbTcpIpTransitioning     bool   `json:"-" bson:"-"` // true while ADB daemon is restarting due to adb tcpip/usb switch; suppresses false device resets
 }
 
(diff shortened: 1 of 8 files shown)


shamanec:
@mogilevich I did some experimenting for this with a separate GADS client cli that authenticates and goes through the hub and it seems to be working fine, I will do a bit more testing and let you know when its ready for testing :D

shamanec:
@mogilevich implemented in the latest release [v.5.4.0](https://github.com/shamanec/GADS/releases/tag/v5.4.0)

mogilevich:
thanks!

PR #269, opened 2026-03-20
Feat(hub): Add HTTP device lease locks and tighten session handling

aloneik:
### Goal
Introduce a reliable HTTP-based device lease lock mechanism (independent from WebSocket heartbeat) to prevent concurrent access conflicts and keep device usage state consistent across Hub flows.

### Summary of Changes

- Added new HTTP endpoints to lock and unlock devices with a lease timeout.
- Extended device runtime state with HTTP lease expiration tracking.
- Added shared helpers to:
     - detect active HTTP/manual locks,
     - clear expired leases,
     - centralize lock field cleanup.
- Updated device in-use logic across critical paths:
     - device proxy access checks,
- WebSocket in-use flow,
- Appium Grid cleanup transitions,
- SSE availability/in-use status,
      - admin forced device release.
- Added session payload validation for Appium session creation (requires clientSecret capability when no auth token is provided).
- Added test coverage for conflict behavior when another user holds an active HTTP lease.
- Included minor provider-side reliability fixes (error/log formatting).

### DIFF
--- a/common/models/models.go
+++ b/common/models/models.go
@@ -208,9 +208,10 @@ type LocalHubDevice struct {
 	InUseTS                  int64    `json:"in_use_ts"`
 	AppiumNewCommandTimeout  int64    `json:"appium_new_command_timeout"`
 	IsAvailableForAutomation bool     `json:"is_available_for_automation"`
-	Available                bool     `json:"available" bson:"-"` // if device is currently available - not only connected, but setup completed
-	InUseWSConnection        net.Conn `json:"-" bson:"-"`         // stores the ws connection made when device is in use to send data from different sources
-	LastActionTS             int64    `json:"-" bson:"-"`         // Timestamp of when was the last time an action was performed via the UI through the proxy to the provider
+	Available                bool     `json:"available" bson:"-"`              // if device is currently available - not only connected, but setup completed
+	InUseWSConnection        net.Conn `json:"-" bson:"-"`                      // stores the ws connection made when device is in use to send data from different sources
+	InUseHTTPExpiresAt       int64    `json:"in_use_http_expires_at" bson:"-"` // timestamp until which the device is considered in use based on HTTP requests - this is used to handle the case when a user has a session but didn't connect via WS for some reason, so we can still consider the device in use until this timestamp expires based on the last HTTP request made by the user for this device
+	LastActionTS             int64    `json:"-" bson:"-"`                      // Timestamp of when was the last time an action was performed via the UI through the proxy to the provider
 }
 
 type DeviceStreamSettings struct {
@@ -235,6 +236,10 @@ type UpdateStreamSettings struct {
 type DeviceInUseMessage struct {
 	Type    string      `json:"type"`
 	Payload interface{} `json:"payload"`
+}
+
+type DeviceLockRequest struct {
+	TimeoutMS int64 `json:"timeout_ms"`
 }
 
 type DBFile struct {
--- a/hub/router/appiumgrid.go
+++ b/hub/router/appiumgrid.go
@@ -257,10 +257,11 @@ func AppiumGridMiddleware() gin.HandlerFunc {
 				foundDevice.IsRunningAutomation = false
 				if resp.StatusCode != http.StatusInternalServerError {
 					// Only clear user info if no manual session is active
-					if foundDevice.InUseWSConnection == nil {
+					if !hasActiveManualLockLocked(foundDevice, time.Now().UnixMilli()) {
 						foundDevice.InUseBy = ""
 						foundDevice.InUseByTenant = ""
 						foundDevice.InUseTS = 0
+						foundDevice.InUseHTTPExpiresAt = 0
 					}
 				}
 				devices.HubDevicesData.Mu.Unlock()
@@ -275,10 +276,11 @@ func AppiumGridMiddleware() gin.HandlerFunc {
 							foundDevice.SessionID = ""
 							foundDevice.IsRunningAutomation = false
 							// Only clear user info if no manual session is active
-							if foundDevice.InUseWSConnection == nil {
+							if !hasActiveManualLockLocked(foundDevice, time.Now().UnixMilli()) {
 								foundDevice.InUseBy = ""
 								foundDevice.InUseByTenant = ""
 								foundDevice.InUseTS = 0
+								foundDevice.InUseHTTPExpiresAt = 0
 							}
 						}
 						devices.HubDevicesData.Mu.Unlock()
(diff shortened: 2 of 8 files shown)


docs/docs.go
shamanec, line 1: Please do not commit swagger documentation files, they get generated automatically by the release action :P

shamanec:
hey @aloneik  sorry for the delay, I wanted to finish up some other stuff before checking this out. Before I actually try to comprehend this PR Imma need some information - whats the issue you encounter that requires this enhancement? I also noticed there are two different endpoints introduced - `/lock` and `/unlock` - where and how do you imagine these should be used from the web UI?

aloneik:
> where and how do you imagine these should be used from the web UI?

Hi @shamanec 
This change is for Jenkins/GitLab CI. Each device has a dedicated node/runner, and the pipeline must block user access via the web UI while the device is in use.

shamanec:
> > where and how do you imagine these should be used from the web UI?
> 
> Hi @shamanec This change is for Jenkins/GitLab CI. Each device has a dedicated node/runner, and the pipeline must block user access via the web UI while the device is in use.

Dedicated node/runner in which aspect? Appium test execution or what? If its Appium aren't you using the integrated GADS Appium Grid which should be taking care of this? In general I dont feel comfortable implementing stuff tightly related to a particular person setup 😅 I will check it out though, if it doesn't create issues for normal usage we might add it 🤷‍♂️

aloneik:
> > > where and how do you imagine these should be used from the web UI?
> > 
> > 
> > Hi @shamanec This change is for Jenkins/GitLab CI. Each device has a dedicated node/runner, and the pipeline must block user access via the web UI while the device is in use.
> 
> Dedicated node/runner in which aspect? Appium test execution or what? If its Appium aren't you using the integrated GADS Appium Grid which should be taking care of this? In general I dont feel comfortable implementing stuff tightly related to a particular person setup 😅 I will check it out though, if it doesn't create issues for normal usage we might add it 🤷‍♂️

Thanks for checking. By dedicated node/runner, I mean a CI agent assigned to a specific physical device, not the integrated Appium Grid. In our setup, a job targets a specific device directly, so the goal is to temporarily lock/reserve that device while the pipeline is running and prevent access from the web UI at the same time. This is not specific to my setup only — STF-based frameworks usually expose HTTP endpoints to lock/unlock devices for exactly this kind of CI coordination. Of course, I agree it only makes sense if it can be added as a generic mechanism without affecting normal usage.

aloneik:
For CI pipelines, I’m not sure how the device lock timeout should be handled, since it’s hard to know in advance how long the required actions on the device will take.
So, an endpoint to extend the lock timeout might also be needed.

shamanec:
I am just trying to understand the use case. What exactly are those runners doing against the devices? Are they running Appium tests or its some kind of scripts that interact with the device APIs from the remote control? Whats the expectation when you try to lock a device that is already being in use from the remote control - should it kick out the user?  What if the device is being used by the same user that you authenticate with against the lock/unlock endpoints? There are too many variables in the current setup that might conflict with the target implementation 😅

aloneik:
I'll try to explain step-by-step

> What exactly are those runners doing against the devices?

They run non-Appium CI jobs against specific devices — mainly executing adb commands on Android devices or running browser-based tests via Puppeteer.

> Are they running Appium tests or its some kind of scripts that interact with the device APIs from the remote control?

The latter. They are custom scripts/jobs interacting with the device directly, not Appium tests.

> Whats the expectation when you try to lock a device that is already being in use from the remote control - should it kick out the user?

That’s still an open question, but I would not expect a CI pipeline to forcibly take over a device that is already in use.
From my perspective, locking should mainly prevent new access while the CI job is running, rather than interrupt an existing user session.

> What if the device is being used by the same user that you authenticate with against the lock/unlock endpoints?

In that case, I would expect the current behavior to apply, where the same user can reuse the existing control session in the web UI.

More generally, I think user/session management for remote control is somewhat outside the scope of this feature. For example, in a CI setup I would normally expect each pipeline or automation flow to use its own dedicated user/account. For my use case, that would be sufficient.

shamanec:
These are important details to take into consideration though. When a CI job wants to take over a device to run against it - when the device is already free, its good to go. But if the device is already in use we have two separate cases:
1. The device is in use by the same user that is being used by the CI job - in this case it makes sense to allow the CI runner to take over without kicking out the user
2. The more important one is when the device is being user by a different user than the CI job - in this case we have two options - kick out the current user and allow the CI runner to take over or fail the CI job - both options sound kind of worrying though :D But letting the CI runner use a device while someone else is remotely controlling it is definitely not a great idea

In general maybe this should be somehow similar to the way Appium test execution works - if a device is in use by another user - we do not allow the device to be picked up for automation, if a device is in use by the same user - we do allow it to be picked up for automation. So something along the line of:

- You call `/lock` with user A - if the device is free or in use by user A - `/lock` succeeds
- You call `/lock` with user A - if the device is in use by user B - `/lock` fails
- You call `/lock` with user A(admin) - if the device is in use by user B - `/lock` succeeds and user B is kicked out from remote control
- You call `/unlock` with user A - if the device is free or in use by user A - `/unlock `succeeds
- You call `/unlock` with user A - if the device is in use by user B - `/unlock` fails
- You call `/unlock` with user A(admin) - if the device is in use by user B - `/unlock` succeeds and user B is kicked out from remote control

So essentially for your CI you will be using preferably an admin user and the CI runners will be able to steal devices from anyone because CI runners are with higher priority or you will have a normal user and CI runs will fail against devices that are being used by someone else
Do my ramblings make any sense? 🤣 

In general the LocalHubDevice requires some refactoring for better extendability and maintainability using reference functions because right now this is just patching over patching over patching code and its become a huge mess :(

aloneik:
Yes, that makes sense to me. The Appium-like ownership rules sound like a good model here. For my use case, it would be enough if /lock succeeded only when the device is free or already in use by the same user, and failed when the device is in use by another user. I do not expect CI to run against a device while another user is remotely controlling it. Same-user reuse also makes sense. Admin takeover could be useful as a separate override mechanism, but it is not strictly required for my immediate case. My main goal is a generic lock/reservation mechanism that CI can use safely, not something specific to my setup. And yes, I understand that if LocalHubDevice needs refactoring first, that is probably the right way to approach it.

shamanec:
@aloneik I refactored the LocalHubDevice a bit and added the lock/unlock endpoints myself with another PR, sorry for hijacking your effort a bit :D do you want a beta release to test them out for your use case?

aloneik:
Thank you! I would like to have a beta release.

shamanec:
[Here](https://github.com/shamanec/GADS/releases/tag/v5.3.0-test-release) it is, note that since the endpoints are going through the Hub they are authenticated (in your PR they were as well, so I assume you have this handled). You could use the Authentication header but the `/lock` and `/unlock` endpoints support the token as a url param as well. The `/lock` also has url param for ttl `ttl_minutes`. So essentially you can do `{hub_address}/devices/control/{udid}/lock?ttl_minutes=10&token={authtoken}` and `{hub_address}/devices/control/{udid}/unlock?token={authtoken}`. Let me know if this works for you and if you find any issues because it was a lot of changes :D

aloneik:
Thanks. I'll try this beta version and let you know if I encounter any problems.

aloneik:
I have noticed a scenario where the locking mechanism works questionably.
When a device is Disabled, but it's possible to book it.
image

Do you think this behavior is correct?

shamanec:
> I have noticed a scenario where the locking mechanism works questionably. When a device is Disabled, but it's possible to book it. image
> 
> Do you think this behavior is correct?

well its a bit suspicious :D I can probably handle this just in case, although I don't see a reason why someone will try to book a device that is not used 😅

shamanec:
I also found another issue that I introduced with these changes but its fixed in my pending PR. If you want you can find me on Discord through the project server and we can discuss stuff less async there :)

PR #272, opened 2026-03-28
Refactor LocalHubDevice and add additional endpoints for reserving and releasing devices

shamanec:
* Move LocalHubDevice model to the hub package
* Implement better device store with improved mutex mechanisms instead of the original devices map
* Add receiver methods for the LocalHubDevice with proper mutex handling for the device updates
* Reorganize the device handling code to use the new methods - cleaner and more readable
* Add /lock and /unlock endpoints for reserving and releasing device from CI or other external sources

### DIFF
--- a/hub/devices/hub_device.go
+++ b/hub/devices/hub_device.go
@@ -77,11 +77,7 @@ func (d *LocalHubDevice) ReleaseLockIfNotHeld() {
 	if d.HasUISession() || d.HasActiveLease() {
 		return
 	}
-	d.InUseBy = ""
-	d.InUseByTenant = ""
-	d.InUseTS = 0
-	d.LockSource = ""
-	d.LeaseExpiresAt = 0
+	d.ReleaseLock()
 }
 
 // IsLocked reports whether the device is currently locked by any means:
--- a/hub/devices/hub_device.go
+++ b/hub/devices/hub_device.go
@@ -25,22 +25,22 @@ const (
 // LocalHubDevice represents a device as tracked by the hub at runtime.
 // All field access must be protected by Mu.
 type LocalHubDevice struct {
-	Mu                       sync.RWMutex `json:"-" bson:"-"` // protects this device's fields
+	Mu                       sync.RWMutex  `json:"-" bson:"-"` // protects this device's fields
 	Device                   models.Device `json:"info"`
-	SessionID                string       `json:"-"`
-	IsRunningAutomation      bool         `json:"is_running_automation"`
-	LastAutomationActionTS   int64        `json:"last_automation_action_ts"`
-	InUse                    bool         `json:"in_use"`
-	InUseBy                  string       `json:"in_use_by"`
-	InUseByTenant            string       `json:"in_use_by_tenant"`
-	InUseTS                  int64        `json:"in_use_ts"`
-	LockSource               string       `json:"lock_source" bson:"-"` // "ui", "api", or ""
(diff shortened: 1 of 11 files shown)


hub/router/routes.go
```diff
@@ -490,192 +489,103 @@ func DeviceInUseWS(c *gin.Context) {
 	}
 
 	// Verify if the device is already in use by another user
-	devices.HubDevicesData.Mu.Lock()
-	device, exists := devices.HubDevicesData.Devices[udid]
+	device, exists := devices.HubDeviceStore.Get(udid)
```
gifflet, line 483: ### Security — High

**Context**: A few lines above this block, in `DeviceInUseWS`, token extraction uses `auth.ExtractTokenFromBearer`. If the token is malformed or missing the `"Bearer "` prefix, extraction fails silently and `username`/`userTenant` remain empty strings — execution reaches this device lookup as an anonymous user.

**Suggested fix**: Reuse `getClaimsFromRequest` (already defined in this file), which handles both `Authorization` header and `?token=` query param, and returns `nil` on any error:

```go
claims := getClaimsFromRequest(c)
if claims == nil || claims.Username == "" {
    c.Status(http.StatusUnauthorized)
    return
}
username = claims.Username
userTenant = claims.Tenant
```

This eliminates the divergence between `DeviceInUseWS` and `LockDevice`/`UnlockDevice` token extraction logic and closes the silent auth bypass for malformed tokens.

hub/devices/hub_device.go
```diff
@@ -0,0 +1,136 @@
...
+	d.InUseBy = ""
+	d.InUseByTenant = ""
+	d.InUseTS = 0
+	d.LockSource = ""
+	d.LeaseExpiresAt = 0
+}
+
+// ReleaseLockIfNotHeld clears the lock only when no UI WebSocket and no active API lease are present.
+// Used by automation cleanup to avoid releasing a manually-held device.
+func (d *LocalHubDevice) ReleaseLockIfNotHeld() {
```
gifflet, line 76: ### Simplicity (KISS) — Medium

`ReleaseLockIfNotHeld` duplicates the same 5 field-clearing lines that already exist in `ReleaseLock`. Since `HasUISession() == false` implies `InUseWSConnection == nil`, calling `ReleaseLock()` here will not close any connection — the behavior is identical.

**Suggested fix**:
```go
// ReleaseLockIfNotHeld clears the lock only when it is not actively held
// by a UI or API session. InUseWSConnection is nil when HasUISession() is
// false, so ReleaseLock will not close any connection.
func (d *LocalHubDevice) ReleaseLockIfNotHeld() {
    if d.HasUISession() || d.HasActiveLease() {
        return
    }
    d.ReleaseLock()
}
```

When new lock fields are added in the future, only `ReleaseLock` needs to be updated.

gifflet (changes requested):
I've reviewed the changes and left some comments inline with suggestions for improvement.

hub/router/routes.go
shamanec: Done [here](https://github.com/shamanec/GADS/pull/272/commits/5bf335fdf1bb3640e1804b6086915be2323d155e)

hub/devices/hub_device.go
shamanec: Done [here](https://github.com/shamanec/GADS/pull/272/commits/5bf335fdf1bb3640e1804b6086915be2323d155e)

gifflet (approved):

PR #273, opened 2026-03-30
Refactor provider device code: split Device struct, introduce PlatformDevice interface

shamanec:
This is a huge refactor with the help of CC to streamline and improve the codebase - specifically the provider devices package and the device models

 - Split models.Device into models.DBDevice (persistent fields) and RuntimeState (provider-only runtime fields). The old struct mixed database records 
  with runtime state like ports, contexts, mutexes, and go-ios objects, making it unclear what gets persisted vs what lives only in memory.             
  - Introduce PlatformDevice interface and per-OS concrete types (AndroidDevice, IOSDevice, TizenDevice, WebOSDevice). Eliminates switch/case on        
  device.OS throughout the codebase in favor of polymorphic method dispatch.                                                                            
  - Replace global DBDeviceMap + manual mutex with DeviceStore. Encapsulates concurrent access with RWMutex and snapshot-based iteration, reducing lock
  contention.                                                                                                                                           
  - Slim down provider-to-hub sync payload from the full Device struct to ProviderDeviceSync (4 fields: UDID, Host, Connected, ProviderState). The hub
  independently reads DB fields from MongoDB, so sending them every second was redundant.                                                               
  - Extract shared Appium setup into appium.go. All platforms now use setupAppiumForDevice() via the AppiumCapabilities() interface method.
  - Move hub runtime fields (Host, Connected, ProviderState, LastUpdatedTimestamp) from the embedded Device to top-level fields on LocalHubDevice where 
  they logically belong.                                                                                                                                
  - Fix WebSocket auth regression in getClaimsFromRequest. The helper wasn't stripping the "Bearer " prefix from ?token= query params, causing the
  device in-use WebSocket to silently fail authentication.                                                                                              
  - Remove unused health.go. checkAppiumSession and createAppiumSession were dead code never called from any path.                                                                         
   
  Breaking changes                                                                                                                                      
                  
  - Provider /devices and /provider-data endpoints now return []ProviderDeviceSync instead of full device data. External consumers (if any) that        
  depended on the full struct from these provider-side endpoints will need updating.
  - Device control endpoints now return 404 when the device UDID is not found (previously could nil-pointer panic).


provider/devices/runtime.go
```diff
@@ -0,0 +1,152 @@
...
+func (r *RuntimeState) SetProviderState(state string)                { r.ProviderState = state }
+func (r *RuntimeState) IsConnected() bool                            { return r.Connected }
+func (r *RuntimeState) SetConnected(connected bool)                  { r.Connected = connected }
+func (r *RuntimeState) GetHost() string                              { return r.Host }
+func (r *RuntimeState) SetHost(host string)                          { r.Host = host }
+func (r *RuntimeState) GetLogger() models.CustomLogger               { return r.Logger }
+func (r *RuntimeState) GetContext() context.Context                  { return r.Context }
+func (r *RuntimeState) GetAppiumPort() string                        { return r.AppiumPort }
+func (r *RuntimeState) SetAppiumPort(port string)                    { r.AppiumPort = port }
+func (r *RuntimeState) GetStreamPort() string                        { return "" } // overridden by Android/iOS
```
gifflet: ### Interface Design — High

**Suggestion:**
`GetStreamPort()` returning `""` from `RuntimeState` is a design lie. The `// overridden by Android/iOS` comment signals the implementation is fighting Go's type system instead of using it.

TV devices (Tizen, WebOS) don't have stream ports, so they shouldn't be forced to implement this method. Remove `GetStreamPort()` from `PlatformDevice` and from `RuntimeState` entirely, and move it to the existing `RemoteControllable` sub-interface. Callers that need it should type-assert:

```go
if rc, ok := platDev.(devices.RemoteControllable); ok {
    port := rc.GetStreamPort()
}
```

**Rationale:** Any caller that forgets to type-assert gets a silent empty-string result instead of a compile-time error. The current approach makes the absence of a stream port a runtime convention rather than a type-system guarantee.

provider/devices/ios.go
```diff
@@ -381,156 +574,127 @@ func goIosDeviceWithRsdProvider(device *models.Device) error {
...
 	}
-	newEntry, err := ios.GetDeviceWithAddress(device.UDID, device.GoIOSTunnel.Address, rsdProvider)
-	newEntry.UserspaceTUN = device.GoIOSDeviceEntry.UserspaceTUN
-	newEntry.UserspaceTUNPort = device.GoIOSDeviceEntry.UserspaceTUNPort
-	device.GoIOSDeviceEntry = newEntry
+	newEntry, err := ios.GetDeviceWithAddress(d.GetUDID(), d.GoIOSTunnel.Address, rsdProvider)
+	newEntry.UserspaceTUN = d.GoIOSDeviceEntry.UserspaceTUN
+	newEntry.UserspaceTUNPort = d.GoIOSDeviceEntry.UserspaceTUNPort
+	d.GoIOSDeviceEntry = newEntry
+	d.GoIOSDeviceEntry = newEntry
```
gifflet: ### Correctness — High

Two bugs on these lines:

1. **Error checked after result is used** — if `GetDeviceWithAddress` returns an error, `newEntry` is a zero-value `ios.DeviceEntry` and the device entry gets overwritten with garbage before the error is caught.
2. **Duplicate assignment** — `d.GoIOSDeviceEntry = newEntry` appears twice (lines 580–581), copy-paste artifact.

**Fix:**
```go
newEntry, err := ios.GetDeviceWithAddress(d.GetUDID(), d.GoIOSTunnel.Address, rsdProvider)
if err != nil {
    return err
}
newEntry.UserspaceTUN = d.GoIOSDeviceEntry.UserspaceTUN
newEntry.UserspaceTUNPort = d.GoIOSDeviceEntry.UserspaceTUNPort
d.GoIOSDeviceEntry = newEntry
```

provider/devices/runtime.go
```diff
@@ -0,0 +1,152 @@
...
+
+	// Infrastructure
+	Context          context.Context
+	CtxCancel        context.CancelFunc
+	Mutex            sync.Mutex
+	SetupMutex       sync.Mutex
+	Logger           models.CustomLogger
+	SemVer           *semver.Version
+	InitialSetupDone bool
+	AppiumReadyChan  chan bool
```
gifflet: ### Correctness — Medium

`AppiumReadyChan` is dead state — it is never initialized, sent to, or received from anywhere in the codebase. `IOSDevice` has its own `WdaReadyChan` which is properly initialized and used. Appium readiness in `appium.go` uses a polling loop with `GetIsAppiumUp()`, not a channel.

A `nil` channel blocks forever if accidentally used in a `select`, making this a latent bug. Remove the field entirely.

provider/devices/appium.go
```diff
@@ -0,0 +1,217 @@
...
+		d.Reset("Failed to allocate free Appium port for device.")
+		return err
+	}
+	d.SetAppiumPort(appiumPort)
+
+	caps := d.AppiumCapabilities()
+	go startAppium(d, caps)
+
+	timeout := time.After(30 * time.Second)
+	tick := time.Tick(200 * time.Millisecond)
```
gifflet: ### Resource Leak — Medium

`time.Tick` creates a ticker whose underlying goroutine can never be garbage collected. Every call to `setupAppiumForDevice` leaks a timer goroutine. On a long-running provider with many device reconnects, this accumulates.

**Fix:**
```go
ticker := time.NewTicker(200 * time.Millisecond)
defer ticker.Stop()
// Replace <-tick with <-ticker.C throughout the select
```

hub/router/routes.go
```diff
@@ -930,7 +926,13 @@ func getClaimsFromRequest(c *gin.Context) *auth.JWTClaims {
...
 		tokenString = t
 	} else if raw := c.Query("token"); raw != "" {
-		tokenString = raw
+		// The query param may arrive as "Bearer <token>" (e.g. from WebSocket URLs).
+		// Try to strip the prefix; if it's a bare token, use it as-is.
+		if t, err := auth.ExtractTokenFromBearer(raw); err == nil {
+			tokenString = t
+		} else {
+			tokenString = raw
+		}
```
gifflet: ### Correctness — Medium

`getClaimsFromRequest` correctly handles both `"Bearer <token>"` and bare tokens from `?token=`. 

However, `AuthMiddleware` in `hub/auth/auth.go` still only handles the `"Bearer "` prefix path — a bare token passed via `?token=` to any endpoint using `AuthMiddleware` will fall through to `AbortUnauthorized`. The fix here is correct but the same logic should be applied to `AuthMiddleware` for consistency, otherwise the two auth code paths diverge in behavior for `?token=` params.

provider/devices/platform.go
```diff
@@ -0,0 +1,105 @@
...
+import (
+	"context"
+
+	"GADS/common/models"
+)
+
+// PlatformDevice is the interface that each OS-specific device type implements.
+// It provides a unified API for device lifecycle and app management,
+// eliminating the need for switch/case on device.OS throughout the codebase.
+type PlatformDevice interface {
```
gifflet, line 21: ### Interface Size / ISP — Medium

`PlatformDevice` has 40+ methods combining lifecycle, app management, 25+ getters/setters, Appium infrastructure, and streaming configuration. TV devices (Tizen, WebOS) must implement stream-related methods that don't apply to them.

The getters/setters exist because `RuntimeState` fields need to be mutated through the interface — this is a sign the interface is being used as a substitute for package-level access rather than as a behavioral contract.

Suggestions for a follow-up refactor:
1. Move all stream-related setters/getters into `RemoteControllable` (already exists as a sub-interface — good start)
2. Move Appium infrastructure methods (`SetAppiumUp`, `SetAppiumLastPingTS`, `SetHasAppiumSession`, etc.) to a package-internal interface or just pass `*RuntimeState` directly to `setupAppiumForDevice` since it's in the same package
3. Remove `GetStreamPort()` from `PlatformDevice` (see separate comment on `runtime.go:78`)

This doesn't have to block the PR but is worth tracking for the next cleanup cycle.

gifflet (changes requested):
I've reviewed the changes and left some inline comments with suggestions for improvement.

provider/devices/appium.go
shamanec: Fixed [here](https://github.com/shamanec/GADS/pull/273/commits/a9ac73c879ab39d8486224a4392662ada2718a7d)

provider/devices/runtime.go
shamanec: Fixed [here](https://github.com/shamanec/GADS/pull/273/commits/a9ac73c879ab39d8486224a4392662ada2718a7d)

provider/devices/ios.go
shamanec: Fixed [here](https://github.com/shamanec/GADS/pull/273/commits/a9ac73c879ab39d8486224a4392662ada2718a7d)

provider/devices/runtime.go
shamanec: Done [here](https://github.com/shamanec/GADS/pull/273/commits/0c3da5c65fed2f6676b51500bd5d7ff73b7ae31a), I also moved other RC related code to the RemoteControllable interface as well

provider/devices/platform.go
shamanec: Some of those were changed [here](https://github.com/shamanec/GADS/pull/273/commits/0c3da5c65fed2f6676b51500bd5d7ff73b7ae31a), I will keep this in mind for future refactoring

hub/router/routes.go
shamanec: Done [here](https://github.com/shamanec/GADS/pull/273/commits/481d40a756d250a1ae9653c82ee2aff588055139)

gifflet (approved):

PR #278, opened 2026-04-03
Add Android local debug

shamanec:
Add a new CLI option `adb-client` that allows you to connect a remotely controlled Android device over adb tunnel to your local PC for development and debugging. Readme has been updated with relevant information on usage

**Note:**
You can use the `adb-client` like this - `./GADS adb-client --hub={hub url} --username={GADS username} --password={GADS password} --udid={device udid}`, e.g. `./GADS adb-client --hub=http://192.168.1.24:10000 --username=admin --password=password --udid=123ABC`. You can only connect devices that are currently being remotely controlled by you.

### DIFF
--- a/client/adb/adb.go
+++ b/client/adb/adb.go
@@ -10,6 +10,7 @@
 package adb
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
@@ -18,6 +19,7 @@ import (
 	"net"
 	"net/http"
 	"net/url"
+	"os"
 	"os/exec"
 	"os/signal"
 	"strings"
@@ -35,6 +37,12 @@ func Start(flags *pflag.FlagSet) {
 	udid, _ := flags.GetString("udid")
 	username, _ := flags.GetString("username")
 	password, _ := flags.GetString("password")
+	if password == "" {
+		password = os.Getenv("GADS_PASSWORD")
+	}
+	if password == "" {
+		log.Fatal("password is required: use --password flag or GADS_PASSWORD env var")
+	}
 	localPort, _ := flags.GetInt("port")
 
 	hub = strings.TrimRight(hub, "/")
@@ -197,10 +205,13 @@ func wsErrorToMessage(err error) error {
 }
 
 func authenticate(hub, username, password string) (string, error) {
-	body := fmt.Sprintf(`{"username":"%s","password":"%s"}`, username, password)
-	resp, err := http.Post(hub+"/authenticate", "application/json", strings.NewReader(body))
+	payload, err := json.Marshal(map[string]string{"username": username, "password": password})
 	if err != nil {
-		return "", err
+		return "", fmt.Errorf("failed to encode credentials: %w", err)
+	}
+	resp, err := http.Post(hub+"/authenticate", "application/json", bytes.NewReader(payload))
+	if err != nil {
+		return "", fmt.Errorf("Failed to post to authenticate endpoint - %w", err)
 	}
 	defer resp.Body.Close()
 
--- a/client/adb/adb.go
+++ b/client/adb/adb.go
@@ -1,0 +1,225 @@
+/*
+ * This file is part of GADS.
+ *
+ * Copyright (c) 2022-2025 Nikola Shabanov
+ *
+ * This source code is licensed under the GNU Affero General Public License v3.0.
+ * You may obtain a copy of the license at https://www.gnu.org/licenses/agpl-3.0.html
+ */
+
+package adb
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"log"
+	"net"
+	"net/http"
+	"net/url"
+	"os/exec"
+	"os/signal"
+	"strings"
+	"syscall"
+	"time"
+
+	"github.com/gobwas/ws"
+	"github.com/spf13/pflag"
+)
+
+func Start(flags *pflag.FlagSet) {
+	log.SetFlags(log.Ldate | log.Ltime)
+
+	hub, _ := flags.GetString("hub")
+	udid, _ := flags.GetString("udid")
+	username, _ := flags.GetString("username")
+	password, _ := flags.GetString("password")
+	localPort, _ := flags.GetInt("port")
+
+	hub = strings.TrimRight(hub, "/")
+
+	// Authenticate
+	token, err := authenticate(hub, username, password)
+	if err != nil {
+		log.Fatalf("Authentication failed: %v", err)
+	}
+	log.Println("Authenticated successfully")
+
+	// Open local TCP listener
+	listenAddr := fmt.Sprintf("localhost:%d", localPort)
+	ln, err := net.Listen("tcp", listenAddr)
+	if err != nil {
+		log.Fatalf("Failed to listen on %s: %v", listenAddr, err)
+	}
+
+	actualPort := ln.Addr().(*net.TCPAddr).Port
+	adbAddr := fmt.Sprintf("localhost:%d", actualPort)
+
+	// Handle shutdown via signal or tunnel loss
+	ctx, cancel := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
+	defer cancel()
+
+	// Build WebSocket URL
+	wsScheme := "ws"
+	if strings.HasPrefix(hub, "https") {
+		wsScheme = "wss"
+	}
+	parsedHub, err := url.Parse(hub)
+	if err != nil {
+		log.Fatalf("Invalid hub URL: %v", err)
+	}
+
+	wsURL := url.URL{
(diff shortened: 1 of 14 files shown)


hub/router/adb_tunnel.go
```diff
@@ -0,0 +1,124 @@
...
+	"github.com/gobwas/ws"
+)
+
+func ADBTunnelHandler(c *gin.Context) {
+	udid := c.Param("udid")
+
+	var username string
+	if claims, err := auth.GetClaimsFromRequest(c); err == nil {
+		username = claims.Username
+	}
```
gifflet: ### Security - High

**Issue**: Silent authentication failure leads to misleading error response.

When `GetClaimsFromRequest` fails (missing/invalid/expired token), `username` silently becomes `""`. The check `device.InUseBy != username` will reject correctly by accident — but this returns a misleading **409 Conflict** instead of **401 Unauthorized**. Every other route in the codebase that needs caller identity returns an explicit 401 on claims extraction failure.

**Original Code**:
```go
var username string
if claims, err := auth.GetClaimsFromRequest(c); err == nil {
    username = claims.Username
}
```

**Suggestion**:
```go
claims, err := auth.GetClaimsFromRequest(c)
if err != nil {
    c.JSON(http.StatusUnauthorized, gin.H{"error": "authentication required"})
    return
}
username := claims.Username
```

hub/router/adb_tunnel.go
```diff
@@ -0,0 +1,124 @@
...
+
+	go func() {
+		io.Copy(clientConn, providerConn)
+		done <- struct{}{}
+	}()
+
+	<-done
+	clientConn.Close()
+	providerConn.Close()
+	<-done
```
gifflet: ### Concurrency / Resource Leak - High

**Issue**: Goroutine leak — unbuffered `done` channel with 2 senders, 1 receiver.

The flow:
1. One `io.Copy` goroutine finishes → sends to `done`.
2. Main goroutine receives → closes both connections → `defer` fires (double-close).
3. The second goroutine unblocks due to the connection close → tries to send to `done`.
4. **Nobody is receiving** — the main goroutine is gone. The second goroutine blocks forever.

This leaks one goroutine per tunnel connection and accumulates silently over the process lifetime. The **same pattern** exists in `provider/router/adb_tunnel.go` and `client/adb/adb.go`.

**Fix**: Use a buffered channel so both goroutines can always send without blocking, and remove the redundant manual `Close()` calls (the `defer` at line 77 already handles them):

```go
done := make(chan struct{}, 2)

go func() {
    io.Copy(providerConn, clientConn)
    done <- struct{}{}
}()

go func() {
    io.Copy(clientConn, providerConn)
    done <- struct{}{}
}()

<-done
// Closing both connections unblocks the still-running io.Copy.
// The second goroutine will send to the buffered channel and exit cleanly.
```

Apply the same fix to `provider/router/adb_tunnel.go` and `client/adb/adb.go`.

main.go
```diff
@@ -61,6 +62,25 @@ func main() {
...
+		Use:   "adb-client",
+		Short: "Connect to a remote Android device via ADB tunnel through the hub",
+		Run: func(cmd *cobra.Command, args []string) {
+			adb.Start(cmd.Flags())
+		},
+	}
+	adbClientCmd.Flags().String("hub", "", "Hub URL (e.g. http://localhost:10000)")
+	adbClientCmd.Flags().String("udid", "", "Device UDID to tunnel")
+	adbClientCmd.Flags().String("username", "", "GADS username")
+	adbClientCmd.Flags().String("password", "", "GADS password")
```
gifflet, line 76: ### Security - Medium

**Issue**: Password exposed in process list and shell history.

`--password` is a required flag, so users will always pass it on the command line. CLI flags are visible to all OS users via `ps aux` / `/proc/<pid>/cmdline` and persist in shell history.

**Suggestion**: Support an env var fallback so the password doesn't need to appear on the command line. Change the flag description and make it optional:

```go
adbClientCmd.Flags().String("password", "", "GADS password (or set GADS_PASSWORD env var)")
// Remove: adbClientCmd.MarkFlagRequired("password")
```

Then in `client/adb/adb.go`, resolve it:
```go
password, _ := flags.GetString("password")
if password == "" {
    password = os.Getenv("GADS_PASSWORD")
}
if password == "" {
    log.Fatal("password is required: use --password flag or GADS_PASSWORD env var")
}
```

**Rationale**: This is a developer debug tool but credentials are still sensitive. Env var fallback is the standard pattern for CLI tools handling secrets (e.g., `GITHUB_TOKEN`, `DOCKER_PASSWORD`).

client/adb/adb.go
```diff
@@ -0,0 +1,225 @@
...
+		return fmt.Errorf("device is not available")
+	case strings.Contains(errMsg, "502"):
+		return fmt.Errorf("hub could not connect to provider - check if the provider is running")
+	default:
+		return err
+	}
+}
+
+func authenticate(hub, username, password string) (string, error) {
+	body := fmt.Sprintf(`{"username":"%s","password":"%s"}`, username, password)
```
gifflet: ### Security - Medium

**Issue**: JSON injection via hand-built string in `authenticate()`.

If `username` or `password` contains `"`, `\`, or newlines, the resulting JSON is malformed or can inject additional fields.

**Original**:
```go
body := fmt.Sprintf(`{"username":"%s","password":"%s"}`, username, password)
```

**Suggestion**: Use `encoding/json` which handles all escaping correctly:
```go
payload, err := json.Marshal(map[string]string{"username": username, "password": password})
if err != nil {
    return "", fmt.Errorf("failed to encode credentials: %w", err)
}
resp, err := http.Post(hub+"/authenticate", "application/json", bytes.NewReader(payload))
```

**Rationale**: Hand-built JSON strings break with special characters and are a well-known source of injection bugs. `encoding/json` is the right tool here.

provider/devices/android.go
```diff
@@ -424,6 +443,32 @@ func (d *AndroidDevice) forwardRemoteServer() error {
...
+	if err := cmd.Run(); err != nil {
+		return fmt.Errorf("enableADBTCPMode: Error executing `%s` - %s", cmd.Args, err)
+	}
+	// Wait for adbd to restart and USB to reconnect
+	time.Sleep(2 * time.Second)
+	return nil
+}
+
+func (d *AndroidDevice) forwardADB() error {
+	return d.forwardPort("5555", d.ADBPort)
```
gifflet: ### KISS - Low

**Issue**: Magic string `"5555"` repeated 3 times (lines 452, 459, 469).

**Suggestion**: Define a package-level constant:
```go
const adbTCPPort = "5555"
```

Then use `adbTCPPort` in all three locations. Makes intent explicit and ensures consistency if the port ever needs to change.

gifflet (changes requested):
I've reviewed the changes and left some comments inline with suggestions for improvement.

provider/devices/android.go
shamanec: Done [here](https://github.com/shamanec/GADS/pull/278/commits/b2235e6a36b0c3f71d417953f88be3a19d14cc1a)

client/adb/adb.go
shamanec: Done [here](https://github.com/shamanec/GADS/pull/278/commits/b2235e6a36b0c3f71d417953f88be3a19d14cc1a)

hub/router/adb_tunnel.go
shamanec: The done channel has 2 senders and 2 receivers, there shouldn't be a resource leak:                                                                                                                                                                               
  1. First io.Copy finishes → sends to done → received at line 121                                                      
  2. Lines 122-123 close both connections → unblocks the second io.Copy                                                 
  3. Second io.Copy returns → sends to done → received at line 124                                                      
  4. Function returns, defer cancel() stops the monitor goroutine via ctx.Done()
  
  I only updated the channel to a buffered [one](https://github.com/shamanec/GADS/pull/278/commits/b2235e6a36b0c3f71d417953f88be3a19d14cc1a) but I prefer leaving the conn.Close() everywhere as safety nets, they might be redundant but won't do anything problematic

main.go
shamanec: Done [here](https://github.com/shamanec/GADS/pull/278/commits/b2235e6a36b0c3f71d417953f88be3a19d14cc1a)

hub/router/adb_tunnel.go
shamanec: Done [here](https://github.com/shamanec/GADS/pull/278/commits/e6221d798ac2a22bcd5cc26a465a4a490fc18b6e)

hub/router/adb_tunnel.go
gifflet: The analysis was based on a misread — there are 2 sequential <-done receives in the main goroutine (lines 121 and 124), so both senders are always consumed. No goroutine leak exists. Marking as resolved.

PR #289, opened 2026-04-25
docs: fix macOS WDA IPA packaging and add Auto-Lock requirement

nbarari:
- Replace cp/zip with ditto for IPA packaging on macOS — zip breaks code signing silently, causing installation to fail with signature errors
- Add xcode-select step to ensure xcodebuild uses full Xcode, not CLT
- Clarify that device must be connected during build for free Apple Developer accounts (device UDID must be in provisioning profile)
- Add Auto-Lock disable requirement under iOS device notes — screen lock drops the iOS tunnel connection and causes WDA to crash in a loop
- Add FAQ entry for the WDA Automation Running crash loop pattern

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -10,6 +10,9 @@
 - *When I start my provider WDA is in a install loop and I get the error: `Error accepting new connection accept tcp [::]:56505: use of closed network connection`*
     -   Make sure you've properly signed and created the uploaded [WebDriverAgent](./provider.md#prepare-webdriveragent-file---linux-windows) ipa
     -   In your provider config in GADS UI make sure you've provided proper bundle identifier for WebDriverAgent, e.g. `com.shamanec.WebDriverAgentRunner`
+- *[macOS] WebDriverAgent starts (I see "Automation Running" on the device) but immediately stops and the provider enters a loop, why?*
+    - The most common cause is the device **Auto-Lock** setting. When the screen locks, iOS drops the tunnel connection, which kills the WebDriverAgent process. Set `Settings > Display & Brightness > Auto-Lock` to **Never** on the test device.
+    - If Auto-Lock is already disabled, the IPA may have been packaged incorrectly. On macOS, the IPA **must** be created using `ditto` — using `cp`/`zip` breaks the code signature silently. See the [Build WebDriverAgent IPA](./provider.md#build-webdriveragent-ipa-file-manually-using-xcode) section.
 - *[macOS/Linux/Windows] I have a connected iOS device where WebDriverAgent installation/start up consistently fails, why?*
     - Make sure you've properly signed and created the uploaded [WebDriverAgent](./provider.md#prepare-webdriveragent-file---linux-windows) ipa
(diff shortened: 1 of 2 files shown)


docs/faq.md
```diff
@@ -10,6 +10,9 @@
 - *When I start my provider WDA is in a install loop and I get the error: `Error accepting new connection accept tcp [::]:56505: use of closed network connection`*
     -   Make sure you've properly signed and created the uploaded [WebDriverAgent](./provider.md#prepare-webdriveragent-file---linux-windows) ipa
     -   In your provider config in GADS UI make sure you've provided proper bundle identifier for WebDriverAgent, e.g. `com.shamanec.WebDriverAgentRunner`
+- *[macOS] WebDriverAgent starts (I see "Automation Running" on the device) but immediately stops and the provider enters a loop, why?*
+    - The most common cause is the device **Auto-Lock** setting. When the screen locks, iOS drops the tunnel connection, which kills the WebDriverAgent process. Set `Settings > Display & Brightness > Auto-Lock` to **Never** on the test device.
```
shamanec, line 14: That shouldn't really be dropping the tunnel process but in general it is recommended that devices do not auto-lock, its a device farm after all 😅

docs/provider.md
```diff
@@ -203,17 +203,28 @@ Fork is kept up to date with latest mainstream.
...
-- Copy the `.app` bundle inside the `Payload` folder
-- Compress the `Payload` directory into an archive (.zip file) and give it a new name with `.ipa` appended to the end of the file name.
+- Select `Product > Build For Testing`. This will create a `Products/Debug-iphoneos` folder under the DerivedData directory.  
+   `Example`: **/Users/<username>/Library/Developer/Xcode/DerivedData/WebDriverAgent-<hash>/Build/Products/Debug-iphoneos**
+- Navigate to that folder and package the IPA using `ditto` (macOS only). Using `cp`/`zip` will break the code signature:
+  ```bash
+  cd /Users/<username>/Library/Developer/Xcode/DerivedData/WebDriverAgent-<hash>/Build/Products/Debug-iphoneos
+  mkdir Payload
+  ditto WebDriverAgentRunner-Runner.app Payload/WebDriverAgentRunner-Runner.app
+  ditto -c -k --sequesterRsrc --keepParent Payload ~/Downloads/WebDriverAgent-signed.ipa
```
shamanec, line 223: In general since you are already updating the readme you can also add that its possible to package the ipa entirely by hand without using the terminal - make Payload directory, put the WebDriverAgentRunner-Runner.app into it, compress the Payload folder through the Finder right-click context menu and rename the file to WDA.ipa or whatever, this is what I usually do :)

shamanec (changes requested):
Extend the `ipa` packaging info, other than that - good additions

nbarari:
These changes fix real issues I encountered. Happy to review any follow-up you want to add, but I'd rather not scope-creep this PR further.

shamanec (approved):

shamanec:
sure.

PR #295, opened 2026-05-06
Fix(provider): Auto-reconnect Android devices stuck in ADB offline state

mogilevich:
## Problem

When a USB-connected Android device has a frozen ADB transport, it appears as `offline` in `adb devices` despite being physically connected. GADS did not detect this state, so the device was perpetually stuck with `connected: false` / `provider_state: init` and never recovered without manual intervention.

Manual fix was: `adb -s <udid> reconnect` — this resets the transport and GADS picks up the device automatically on the next polling tick.

## Solution

Extended `getConnectedDevicesAndroid()` to detect `offline` status in `adb devices` output and automatically issue `adb -s <udid> reconnect` with a **30-second cooldown** per device to avoid spamming the command on every 1-second polling tick.

A mutex-protected map tracks the last reconnect attempt per UDID, ensuring the check-and-update is atomic.

## Changes

- `provider/devices/android.go`: replaced the `strings.Contains(line, "device")` filter with an explicit `switch` on ADB status field — `"device"` adds to connected list, `"offline"` triggers reconnect goroutine. `unauthorized` and other statuses are explicitly ignored.

### DIFF
--- a/provider/devices/android.go
+++ b/provider/devices/android.go
@@ -31,6 +31,7 @@ import (
 	"regexp"
 	"slices"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/gobwas/ws"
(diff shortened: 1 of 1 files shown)


shamanec (approved):

mogilevich:
thanks!

PR #296, opened 2026-05-06
fix(provider): Add per-device setup backoff for Android and iOS

mogilevich:
## Problem

When a USB-connected Android device is present in `adb devices` with status `device` but all adb shell commands fail immediately with `exit status 1` (e.g. USB debugging reset, unauthorized, or hardware issue), the provider enters a tight retry loop with zero delay:

```
setup starts → adb command fails (exit status 1, instant) → reset →
setup starts → adb command fails (exit status 1, instant) → reset → ...
```

This spawns hundreds of goroutines and child `adb` processes per second per affected device. With multiple such devices simultaneously, the adb server becomes overwhelmed and crashes — disconnecting **all** devices on the provider, including healthy ones.

Observed in production:
- One device: "Failed to disable auto-rotation - exit status 1" — 2617 occurrences over ~6 hours
- Two other devices: ~1 failure/second each
- Eventually: all 24 Android devices simultaneously lost with `signal: killed` on all subprocesses (adb server crash)

> **Note:** this is distinct from the `offline` transport fix in PR #295. That fix handles frozen adb transports. This issue is about devices that appear as `device` but whose commands fail instantly.

## Root cause

No delay between setup retries. When setup fails due to an immediate error (`exit status 1`), the `updateDevices()` 1-second polling loop spawns a new `go platDev.Setup()` goroutine on the very next tick.

## Fix

Added a per-device exponential backoff in `provider/devices/android.go`:

| Consecutive failures | Delay before next attempt |
|---|---|
| 1 | 10s |
| 2 | 20s |
| 3 | 40s |
| 4+ | 60s (cap) |

Errors from context cancellation or `signal: killed` are not counted — those are external resets, not device faults. Healthy devices see zero added latency (backoff is cleared on successful setup).

The shift count is capped at 3 before computing the delay to prevent `int64` overflow for devices with a very large failure count.

## Testing

- Confirmed `go build ./...` passes cleanly
- On a healthy device: setup proceeds immediately with no delay
- On a bad-state device: after the first failure, subsequent attempts are throttled to at most once per 10–60s instead of once per second


provider/devices/android.go
```diff
@@ -58,11 +74,69 @@ func (d *AndroidDevice) GetAndroidIMEPort() string          { return d.AndroidIM
...
 	defer d.SetupMutex.Unlock()
 
+	defer func() {
+		if retErr != nil {
+			recordAndroidSetupFailure(d.GetUDID(), retErr)
+		} else {
+			resetAndroidSetupBackoff(d.GetUDID())
+		}
+	}()
+
```
shamanec, line 87: The whole implementation seems reasonable but a bit overengineered. Why don't you just add two more fields to the `AndroidDevice` struct like
```go
setupBackoffUntil time.Time                                                                                                                            
setupBackoffNext  time.Duration
```
Then remove both `recordAndroidSetupFailure` and `shouldAttemptAndroidSetup` functions, also the `androidSetupBackoffMu`, the `androidSetupBackoffState` type and `androidSetupBackoffs` map. After that here in the Setup() function you can just do something in this context:
```go
d.SetupMutex.Lock()                                                                                                                                    
defer d.SetupMutex.Unlock()                                                                                                                            
                                                                                                                                                             
if time.Now().Before(d.setupBackoffUntil) {                                                                                                            
          return nil
}                                                                                                                                                      
                  
defer func() {                                                                                                                                         
   switch {
    case retErr == nil:                                                                                                                                
              d.setupBackoffNext = 0
              d.setupBackoffUntil = time.Time{}
     case errors.Is(retErr, context.Canceled), errors.Is(retErr, context.DeadlineExceeded):                                                             
              // external cancellation
     default:                                                                                                                                           
              if d.setupBackoffNext == 0 {
                  d.setupBackoffNext = androidSetupBackoffBase                                                                                               
              } else {
                  d.setupBackoffNext = min(d.setupBackoffNext*2, androidSetupBackoffMax)                                                                     
              }
              d.setupBackoffUntil = time.Now().Add(d.setupBackoffNext)                                                                                       
  }                                                                                                                                                  
}()
```
I haven't really tested this but looks reasonable and much simpler and easier to read than the bitshifting implementation, what do you think? Let me know if I am missing something from your code that is not covered here

shamanec: You can maybe even add `setupBackoffUntil` and `setupBackoffNext` to `RuntimeState` instead of `AndroidDevice` struct and use the same implementation for the other PR for the iOS backoff retries? Or just add them both on `AndroidDevice` and `iOSDevice`, not sure if this is applicable for Tizen/WebOS as a problem

shamanec (changes requested):

mogilevich:
Thanks for the feedback! Updated the implementation following your suggestions:

- Removed the global `map` + `sync.Mutex` + three helper functions
- Added `setupBackoffUntil time.Time` and `setupBackoffNext time.Duration` directly to `RuntimeState` — protected by the existing `SetupMutex`, no extra synchronisation needed
- Backoff check is now inline in `Setup()`, right after `SetupMutex.Lock()`
- Switched from string matching to `errors.Is(retErr, context.Canceled)` / `errors.Is(retErr, context.DeadlineExceeded)` for external cancellation detection
- Extended the same pattern to `IOSDevice.Setup()` as you suggested — closed the separate PR #297 in favour of this one

shamanec (approved):

shamanec:
Cool, thanks!

PR #303, opened 2026-05-13
fix(provider): Support iPhone 16 Pro/Max (A18 Pro) by downloading PDI from doronz88/DeveloperDiskImage

mogilevich:
## Problem

iPhone 16 Pro and 16 Pro Max (chip A18 Pro, \`ApChipID=0x8140\`, \`ApBoardID=0xe\`) fail to start WDA on the iOS provider with:

\`\`\`
Failed to mount Developer Disk Image (DDI) for device \`00008140-...\` -
failed to mount DDI: MountImage: could not find identity for identifiers
{BoardId:14 ChipID:33088 ...}: findIdentity: failed to find identity for
ApBoardId 0xe and ApChipId 0x8140
\`\`\`

**Root cause:** go-ios v1.0.202 hardcodes \`ddi-15F31d\` for all iOS 17+ devices. Its \`BuildManifest.plist\` only contains identities up to A16 Bionic (chip \`0x8132\`). The A18 Pro (\`0x8140\`) is absent, so \`MountImage\` fails with a \`findIdentity\` error and GADS resets the device.

## Solution

For iOS 17+, bypass go-ios's hardcoded DDI download and use [doronz88/DeveloperDiskImage](https://github.com/doronz88/DeveloperDiskImage/tree/main/PersonalizedImages/Xcode_iOS_DDI_Personalized), which maintains up-to-date chip identities including A18 Pro.

**Key insight:** The repo ships two generic files (\`Image.dmg\`, \`Image.dmg.trustcache\`), but \`BuildManifest.plist\` references them by versioned names (e.g. \`022-21627-023.dmg\`, \`Firmware/022-21627-023.dmg.trustcache\`). go-ios reads the manifest and looks for files by those exact paths — so the files must be saved under the manifest-specified names.

**Algorithm:**
1. Download \`BuildManifest.plist\` (refreshed every 24h via TTL, atomic replace via \`.tmp\` → rename)
2. Parse it with \`howett.net/plist\` (already an indirect dependency) to extract \`dmgRelPath\` and \`tcRelPath\` from the first \`BuildIdentity\`
3. Download \`Image.dmg\` → save as \`dir/dmgRelPath\`, download \`Image.dmg.trustcache\` → save as \`dir/tcRelPath\` (creating subdirectories as needed)
4. Return \`dir\` — \`imagemounter.MountImage\` finds the files, contacts Apple TSS for a personalized signature, and mounts the PDI

**Other details:**
- Files cached in \`<providerFolder>/devimages/doronz88-pdi/\`
- Package-level mutex prevents concurrent downloads on first provider startup
- For iOS < 17, existing \`imagemounter.DownloadImageFor()\` behaviour is unchanged
- Works cross-platform (macOS, Linux, Windows) — pure HTTP download, no system tools required

## Testing

Verified on a real device fleet:
- ✅ iPhone 16 Pro Max (A18 Pro, \`0x8140\`, UDID \`00008140-0018794802E8401C\`) — DDI mounted, WDA started
- ✅ 24 other devices across iOS versions — no regressions

Log output on success:
\`\`\`
Downloading Image.dmg as 022-21627-023.dmg from doronz88/DeveloperDiskImage
Downloading Image.dmg.trustcache as Firmware/022-21627-023.dmg.trustcache from doronz88/DeveloperDiskImage
Successfully started WebDriverAgent for device 00008140-0018794802E8401C
\`\`\`

### DIFF
--- a/provider/devices/ios.go
+++ b/provider/devices/ios.go
@@ -18,6 +18,7 @@ import (
 	"io"
 	"net/http"
 	"os"
+	"os/exec"
 	"strconv"
 	"strings"
 	"time"
@@ -380,7 +381,13 @@ func (d *IOSDevice) mountDeveloperImage() error {
 			return nil
 		}
 		if strings.Contains(err.Error(), "findIdentity") && d.SemVer.Major() >= 17 {
-			logger.ProviderLogger.LogWarn("ios_device_setup", fmt.Sprintf("Skipping DDI mount for device `%s` (iOS %s): chip not in go-ios identities, DDI not required on iOS 17+", d.GetUDID(), d.SemVer.Original()))
+			logger.ProviderLogger.LogWarn("ios_device_setup", fmt.Sprintf("Chip not in go-ios DDI identities for device `%s` (iOS %s), attempting PDI mount via CoreDevice", d.GetUDID(), d.SemVer.Original()))
+			out, xcrunErr := exec.Command("xcrun", "devicectl", "device", "info", "details", "--device", d.GetUDID()).CombinedOutput()
+			if xcrunErr != nil {
+				logger.ProviderLogger.LogWarn("ios_device_setup", fmt.Sprintf("xcrun devicectl failed for device `%s` - %s: %s", d.GetUDID(), xcrunErr, strings.TrimSpace(string(out))))
+			} else {
+				logger.ProviderLogger.LogInfo("ios_device_setup", fmt.Sprintf("PDI mounted via CoreDevice for device `%s`", d.GetUDID()))
+			}
 			return nil
 		}
 		return fmt.Errorf("failed to mount DDI: %w", err)
--- a/provider/devices/ios.go
+++ b/provider/devices/ios.go
@@ -18,6 +18,7 @@ import (
 	"io"
 	"net/http"
 	"os"
+	"path/filepath"
 	"strconv"
 	"sync"
 	"strings"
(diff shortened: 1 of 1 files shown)


shamanec:
This is not a viable solution because it would work only on macOS and not solve the problem for users on Win/Linux. The correct approach would be to modify the DDI mounting for iOS 17+ to use the maintained DDI files [here](https://github.com/doronz88/DeveloperDiskImage/tree/main/PersonalizedImages/Xcode_iOS_DDI_Personalized)

mogilevich:
Thanks for the feedback =)
Updated the PR to use doronz88/DeveloperDiskImage directly as you suggested:
- Download `BuildManifest.plist` and parse it with `howett.net/plist` to get the versioned file paths go-ios expects
- Download `Image.dmg` / `Image.dmg.trustcache` and save them under the manifest-specified names (e.g. `022-21627-023.dmg`) so `imagemounter.MountImage` finds them
- Pure HTTP download, works cross-platform (macOS, Linux, Windows)

Tested on a real fleet — iPhone 16 Pro Max (A18 Pro, `0x8140`) mounts DDI and starts WDA successfully alongside 24 other devices with no regressions.

shamanec:
Might have been cleaner if we embedded them in the GADS binary but this way we won't have to make a release to update it even if its a bit too much code 😅 GJ! Would you mind contacting me on Discord?

shamanec (approved):

PR #316, opened 2026-07-15
fix(ios): tolerate HangUp failure on iOS 27+ DDI mount

### LINKED ISSUE #315: fix(ios): iOS 27 DDI mount loops on Hangup failure (mount succeeds, cleanup rejected)
## Summary

GADS v5.7.0 fails to set up iOS 27 (beta) devices. The DDI mount **succeeds**, but go-ios's trailing `Hangup` cleanup command fails because iOS 27 closes the `mobile_image_mounter` lockdown connection immediately after the mount. GADS treats this cleanup error as a fatal mount failure and resets the device in a ~12s retry loop, never reaching WDA.

## Device

| Field | Value |
|-------|-------|
| Device | iPhone 14 (iPhone14,7) |
| UDID | `00008110-00040994117A602E` |
| iOS | 27.0 (27A5209h) |
| Xcode | 26.3 (17C529) |
| GADS | v5.7.0 |
| go-ios | v1.0.202 |

## Error

```
Failed to mount Developer Disk Image (DDI) for device `00008110-00040994117A602E` - failed to mount DDI: MountImage: HangUp command failed: Write: only 0 bytes were written instead of 229
```

Repeats every ~12s in an infinite retry loop.

## Root cause

The go-ios personalized DDI mount flow (`personalized_image_mounter.go:67-130`) is:

1. Query identifiers → get nonce → get Apple TSS signature
2. `sendUploadRequest` → stream DMG bytes → `waitForUploadComplete`
3. `mountPersonalizedImage` ← **the actual mount; returns `nil` (succeeds)**
4. `hangUp` ← sends `{"Command":"Hangup"}` over the same lockdown connection

On iOS 27, the device closes the `mobile_image_mounter` lockdown connection immediately after step 3 succeeds. Step 4's write gets 0 bytes → `Write: only 0 bytes were written instead of 229` → wrapped as `MountImage: HangUp command failed`.

GADS's `mountDeveloperImage()` (`provider/devices/ios.go:104`) treats any non-"already mounted" error as fatal and calls `resetWithError`, which resets the device and triggers the retry loop — even though the DDI is already mounted.

This is **not** a lockdown-vs-CoreDevice protocol issue. The `MountImage`/`Personalized` plist command works; only the trailing `Hangup` is rejected. Xcode's `devicectl` uses CoreDevice and never sends `Hangup`, so it doesn't hit this.

Note: the latest go-ios (v1.0.218) adds a `QueryPersonalizationManifest` optimization but still sends the same `Hangup` — so bumping go-ios alone won't fix this.

## Fix

When `imagemounter.MountImage` fails with `"HangUp command failed"`, log a warning and re-connect to verify the DDI is mounted via `ListImages`. If an image is mounted, proceed to WDA. If verification fails, reset as before.

```go
if strings.Contains(err.Error(), "HangUp command failed") {
    logger.ProviderLogger.LogWarn("ios_device_setup", ...)
    return d.verifyDDIMounted()
}
```

`verifyDDIMounted` opens a fresh `NewImageMounter` + `ListImages` on a new lockdown connection, independent of the closed mounter connection.

## Verification

- `go build ./provider/devices/` ✅
- `go vet ./provider/devices/` ✅
- **Live device verification not yet done** — needs the iPhone 14 on iOS 27. The fix is testable by confirming setup proceeds past DDI mount to `setupTunnelIfNeeded()` → WDA install.

## Related

- #309 (go-ios bump for iOS 26.x) — related but distinct; this is an iOS 27-specific hangup-tolerance issue
- go-ios upstream: the `hangUp` call is at `ios/imagemounter/imagemounter.go:180` (v1.0.202)

MaatheusGois:
## Summary

On iOS 27, the device closes the `mobile_image_mounter` lockdown connection immediately after the personalized DDI mount succeeds. The trailing `HangUp` command gets 0 bytes instead of the expected 229-byte response, causing `mountDeveloperImage()` to treat a successful mount as a fatal error and enter an infinite retry loop.

## Fix

When `MountImage` fails with `"HangUp command failed"`, log a warning and verify the DDI is actually mounted by opening a fresh connection and calling `ListImages`. Retry up to 3 times with increasing delay (2s, 4s, 6s) to account for the device needing time to register the mount.

## Root Cause

go-ios's `PersonalizedDeveloperDiskImageMounter.MountImage()` flow:
1. Query identifiers → get nonce → get Apple TSS signature
2. sendUploadRequest → stream DMG bytes → waitForUploadComplete  
3. mountPersonalizedImage ← the actual mount; **returns nil (succeeds)**
4. hangUp ← sends `{"Command":"Hangup"}` over the same lockdown connection

On iOS 27, the device closes the connection after step 3. Step 4's write gets 0 bytes → `MountImage: HangUp command failed`.

## Device

| Field | Value |
|-------|-------|
| Device | iPhone 14 (iPhone14,7) |
| iOS | 27.0 (27A5209h) |
| GADS | v5.7.0 |

## Testing

- `go build ./provider/devices/` ✅
- `go vet ./provider/devices/` ✅  
- Live verified: setup now proceeds past DDI mount to WDA installation on iOS 27

Fixes #315

### DIFF
--- a/provider/devices/ios.go
+++ b/provider/devices/ios.go
@@ -409,8 +409,37 @@ func (d *IOSDevice) mountDeveloperImage() error {
 		if strings.Contains(err.Error(), "already mounted") || strings.Contains(err.Error(), "AlreadyMounted") {
 			return nil
 		}
+		if strings.Contains(err.Error(), "HangUp command failed") {
+			logger.ProviderLogger.LogWarn("ios_device_setup", fmt.Sprintf("DDI mount succeeded but HangUp rejected by device `%s` (likely iOS 27+), verifying mount via ListImages", d.GetUDID()))
+			for attempt := 0; attempt < 3; attempt++ {
+				time.Sleep(time.Duration(2+attempt*2) * time.Second)
+				if verifyErr := d.verifyDDIMounted(); verifyErr != nil {
+					logger.ProviderLogger.LogWarn("ios_device_setup", fmt.Sprintf("DDI verification attempt %d failed for device `%s`: %s", attempt+1, d.GetUDID(), verifyErr))
+					continue
+				}
+				return nil
+			}
+			return fmt.Errorf("failed to mount DDI: %w", err)
+		}
 		return fmt.Errorf("failed to mount DDI: %w", err)
 	}
+	return nil
+}
+
+func (d *IOSDevice) verifyDDIMounted() error {
+	mounter, err := imagemounter.NewImageMounter(d.GoIOSDeviceEntry)
+	if err != nil {
+		return fmt.Errorf("verifyDDIMounted: failed to create image mounter: %w", err)
+	}
+	defer mounter.Close()
+	signatures, err := mounter.ListImages()
+	if err != nil {
+		return fmt.Errorf("verifyDDIMounted: ListImages failed: %w", err)
+	}
+	if len(signatures) == 0 {
+		return fmt.Errorf("verifyDDIMounted: no developer images found on device")
+	}
+	logger.ProviderLogger.LogInfo("ios_device_setup", fmt.Sprintf("DDI verified mounted on device `%s` (%d image(s) found)", d.GetUDID(), len(signatures)))
 	return nil
 }
 


provider/devices/ios.go
```diff
@@ -409,11 +409,40 @@ func (d *IOSDevice) mountDeveloperImage() error {
...
+			for attempt := 0; attempt < 3; attempt++ {
+				time.Sleep(time.Duration(2+attempt*2) * time.Second)
+				if verifyErr := d.verifyDDIMounted(); verifyErr != nil {
+					logger.ProviderLogger.LogWarn("ios_device_setup", fmt.Sprintf("DDI verification attempt %d failed for device `%s`: %s", attempt+1, d.GetUDID(), verifyErr))
+					continue
+				}
+				return nil
+			}
+			return fmt.Errorf("failed to mount DDI: %w", err)
+		}
```
shamanec, line 423:
```suggestion
		if strings.Contains(err.Error(), "HangUp command failed") {
			logger.ProviderLogger.LogWarn("ios_device_setup", fmt.Sprintf("DDI mount succeeded but HangUp rejected by device `%s` (likely iOS 27+), verifying mount via ListImages", d.GetUDID()))
			var verifyErr error
			for attempt := range 3 {
				time.Sleep(time.Duration(2+attempt*2) * time.Second)
				if verifyErr = d.verifyDDIMounted(); verifyErr == nil {
					return nil
				}
				logger.ProviderLogger.LogWarn("ios_device_setup", fmt.Sprintf("DDI verification attempt %d failed for device `%s`: %s", attempt+1, d.GetUDID(), verifyErr))
			}
			return fmt.Errorf("failed to mount DDI: %w", verifyErr)
		}
```

shamanec:
Looks reasonable, you gotta target the develop branch though 👌

PR #332, opened 2026-07-23
Retry stream-port dial to avoid provisioning reset loop

gifflet:
Android device provisioning could get stuck in a reset loop with: `could not update GADS stream settings on device - failed connecting to device <udid> stream port - EOF`

## Root cause
The on-device H264 stream server (`app_process ... H264Server`) takes ~4s to bind its WebSocket port (1991) after being (re)started, but `UpdateStreamSettingsOnDevice` dialed it once, right after startup, with no retry. Losing that race returned an error, which failed `applyStreamConfig` and triggered a full device reset — and the next attempt hit the same race, looping forever.

## Fix
`UpdateStreamSettingsOnDevice` now retries the WebSocket dial for a bounded window (30 × 500 ms = 15 s), cancellable via the device context, instead of failing on the first attempt.

### DIFF
--- a/provider/devices/android.go
+++ b/provider/devices/android.go
@@ -24,6 +24,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
 	"net/url"
 	"os"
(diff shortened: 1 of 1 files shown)


shamanec (approved):

shamanec:
I have a literally 10 year old Android device on Android 7 and I've never seen the server take 4 seconds to fully start, but this retry won't hurt 🤷‍♂️

PR #333, opened 2026-07-24
Stream Android device audio via Opus over WebRTC

gifflet:
Provider/hub side of Android audio streaming. The provider connects to the device's audio WebSocket, encodes the PCM to Opus, and adds an Opus audio track to the Android WebRTC session so the browser gets audio alongside video.


provider/devices/android.go
```diff
@@ -475,6 +489,108 @@ func (d *AndroidDevice) forwardRemoteServer() error {
...
+		d.grantRecordAudioPermission()
+		d.startAudioService()
+		return
+	}
+
+	// Internal device audio (default): needs a MediaProjection token.
+	if d.SemVer != nil && d.SemVer.Major() >= 15 {
+		// On Android 15 (API 35) QPR1 the keyguard stops MediaProjection; disabling it
+		// keeps internal audio capture alive on device farms.
+		d.disableKeyguard()
```
shamanec: Do you think its sensible to do this for all devices in general instead of just when doing audio streaming?

provider/router/android_stream_webrtc.go
```diff
@@ -267,6 +319,41 @@ func (s *AndroidWebRTCSession) writeFrames() {
...
+			logger.ProviderLogger.LogError("webrtc_session", fmt.Sprintf("Error writing audio sample: %v", err))
+			return
+		}
+
+		if frameCount == 1 {
+			logger.ProviderLogger.LogInfo("webrtc_session", fmt.Sprintf("First audio frame written to WebRTC track for device %s (size: %d bytes)", s.device.UDID, len(audioFrame.Data)))
+		}
+		if frameCount%100 == 0 {
+			logger.ProviderLogger.LogDebug("webrtc_session", fmt.Sprintf("Sent audio frame #%d for device %s", frameCount, s.device.UDID))
+		}
```
shamanec, line 351: Won't this spam the provider logs even if is on each 100 frames? Don't we get a loot of frames?

provider/devices/android.go
gifflet: Done [here](https://github.com/shamanec/GADS/pull/333/commits/994e917dfdd85d5e628e784f73c7f2a5353b792d)

provider/router/android_stream_webrtc.go
gifflet: Normally, it shouldn't be a problem if the default log level is `info`. Also, `android_audio_extractor.go` uses the same logging. But if you think this behavior should change, should we update both?

shamanec: Nah, I guess debug level is not the usual log level used and maybe we should expect some spam when using it, if someone complains they have too much noise while using audio streaming maybe we can remove those or reduce the number of logs 🤷‍♂️

shamanec (approved):

PR #336, opened 2026-07-27
Tenant-scoped app library and app install authorization

gifflet:
Uploaded apps are now associated with the uploader's tenant, and the app library only shows each user the apps from their own tenant (admins keep full visibility). The hub also enforces this boundary at install time: installStoredApp requests through the device proxy are rejected when the referenced file belongs to a different tenant, closing the gap where any authenticated user could install any stored app by file ID


shamanec (approved):

shamanec:
I am starting to get really fed up with the tenants. I have high suspicion you are the only one using them and its much easier for me to make people have different deployments for "tenants" instead of constantly having to rework code to accommodate this highly complex while useless to 98% of users functionality

PR #343, opened 2026-07-31
provider: remove stale adb forward rules on device reset

aloneik:
AndroidDevice.Reset() only freed host ports from the in-memory UsedPorts map, but never removed the correspoding `adb forward` rules. Since Setup() is retried automatically (backoff, USB hiccups, transpient failures) and always allocates fresh random hosts ports, every re-setup cycle left the previous cycle's forward rules dangling on the host, causing them to accumulate indefinitely until the providerr process was restarted.

Add removeForwardPort() to run `adb -s <serial> forward --remove tcp:<port> for the stream, IME, remote-server and ADB tunnel ports, and callit from Reset() before the ports are freed and resused.


aloneik:
wrong target branch

PR #345, opened 2026-08-03
Update go-ios dep to v1.2.1

shamanec:
* Rework DDI mounting for iOS
* Migrate from logrus to slog and handle go-ios log capture
* Update hraban/opus to jj11hh/opus

@gifflet release pipeline is broken because of the CGO and opus dependencies, I migrated the audio pipeline to jj11hh/opus because it is pure Go. Unfortunately I wasn't able to test it because the audio pipeline also doesn't work - its made to setup audio streaming only when using H264 streaming server and that server runs on the same port as the audio service - 1992, not sure how you tested it before merging those changes


gifflet:
@shamanec, what exactly is broken in the release pipeline? It seems to be running fine. As you can see, the update-go-ios release was created successfully:
https://github.com/gifflet/GADS/releases/tag/update-go-ios

And here are the corresponding build logs:
https://github.com/gifflet/GADS/actions/runs/30839826069/job/91775274619

shamanec:
Exactly, can you try releasing before this PR? I was not able to build locally without CGO and opus installed via brew? You might also have to update the hub ui commit ref, there is something wrong there as well

gifflet:
@shamanec, we need to update the reference to the latest hub-ui commit. #346

gifflet:
@shamanec , PR #346 merged

shamanec:
@gifflet https://github.com/shamanec/GADS/actions/runs/30874966430/job/91884499823

gifflet:
@shamanec, just create a new tag pointing to commit [c4c0966](https://github.com/shamanec/GADS/commit/c4c09663e88f78bdfff4b60f5ca574ccae15ddd5), and it should work fine (as you can see from this run): https://github.com/gifflet/GADS/actions/runs/30896248293

shamanec:
@gifflet thats what I am trying to say, this PR changed from habran/opus to jj11hh/opus which does not require opus installed on the system and CGO, the build action I referenced was from the develop branch before that change and the build was failing, I believe you missed these dependency requirements

gifflet:
Yeah, in fact. Sorry about that. Two months ago, when I first implemented the audio streaming, I got it working: https://github.com/gifflet/GADS/actions/runs/25697229652. I just forgot to apply the updated workflow to the new audio streaming implementation.

shamanec:
I don't think we need to change the pipeline because it builds fine with jj11hh/opus, but I wasn't able to get the audio streaming working at all to test it due to h264 server and the audio service using the same port 1992 :(

PR #351, opened 2026-08-08
fix(archive): check zip entry paths stay within destination directory

bunlongheng:
The three zip extraction helpers - UnzipInMemory, Unzip, and ExtractZipToDir - join entry names directly to the destination path without checking whether the result stays inside dest. A crafted archive with entry names like ../../../../.ssh/authorized_keys can write files outside the target directory. Added a filepath.Clean prefix check that rejects any entry whose resolved path escapes dest, and tightened MkdirAll permissions from 0777 to 0755.

### DIFF
--- a/common/utils/archive.go
+++ b/common/utils/archive.go
@@ -17,6 +17,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"strings"
 )
 
 func ListFilesInZip(zipData []byte) ([]string, error) {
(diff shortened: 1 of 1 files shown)


shamanec:
This seems kinda reasonable, but I can't understand exactly what kind of a problem it solves? We use those extraction helpers only for installing applications on devices, noone is going to attempt installing a crafted archive with entry names like ` ../../../../.ssh/authorized_keys` as an application🤔Looks like a simple random AI code improvement suggestion after AI analyzed the repo code to me :D

PR #356, opened 2026-08-19
Support optional TURN over TLS (turns:) port

gifflet:
Add a tls_port field to the TURN configuration so the hub and providers can advertise a turns: (TURN over TLS) ICE URL. TLS is optional: turns: is advertised only when tls_port is set (> 0); with tls_port 0 only turn: (UDP/TCP) URLs are generated.

- models.TURNConfig: tls_port field (optional; 0 = TLS disabled)
- hub/provider WebRTC config: append turns: only when tls_port > 0
- UpdateTURNConfig: validate tls_port 0-65535
- default TURN config: tls_port 0
- docs: document the optional TLS Port in the WebRTC TURN server guide
- bump hub-ui submodule (optional TLS Port field in Global Settings)

### DIFF
--- a/hub/router/webrtc.go
+++ b/hub/router/webrtc.go
@@ -14,6 +14,7 @@ import (
 	"GADS/common/db"
 	"GADS/hub/config"
 	"fmt"
+	"log/slog"
 	"net/http"
 
 	"github.com/gin-gonic/gin"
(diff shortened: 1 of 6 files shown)


hub/router/webrtc.go
```diff
@@ -33,27 +33,66 @@ func GetICEConfig(c *gin.Context) {
...
 
+	// Check if TURN is configured and enabled
+	if !turnConfig.Enabled {
+		c.JSON(http.StatusPreconditionFailed, gin.H{
+			"error":   "TURN server not configured",
+			"message": "WebRTC requires TURN server to work in restricted networks. Please configure TURN server in Admin → Global Settings.",
+			"action":  "configure_turn",
+		})
+		return
+	}
```
shamanec: ? Why should TURN be mandatorily configured?

gifflet: Fixed.

```go
turnConfig, err := db.GlobalMongoStore.GetTURNConfig()
	switch {
	case err != nil:
		slog.Warn(fmt.Sprintf("ice-config: could not load TURN config, serving STUN only - %s", err))
	case !turnConfig.Enabled:
		slog.Debug("ice-config: TURN disabled, serving STUN only")
	case turnConfig.Server == "" || turnConfig.SharedSecret == "":
		slog.Warn("ice-config: TURN is enabled but its config is incomplete (missing server or shared secret); serving STUN only")
	default:
```

shamanec (approved):

shamanec:
To be fair this hardly gives any additional security and probably also reduces performance but maybe it makes sense for some super hyper restrictive networks that might not let a normal TURN through

PR #360, opened 2026-09-16
fix(appium): complete device-scoped session cleanup before reuse (#355)

### LINKED ISSUE #355: "Could not create Appium driver" on new session appears often on version 6.0
Hi,

So I'm noticing in version 6.0 that when I execute tests on my iOs device and tests let's say failes and I start a new test again , I quite often get now: "Could not create Appium driver"

then need to restart provider to be able to run new tests

any ideas?

bjelac:
GADS 6.0 released expired device claims immediately after starting an asynchronous Appium DELETE. A subsequent session could overlap the old driver's teardown, and a delayed plugin callback could clear the new session's provider state.

Keep the device reserved until cleanup succeeds and serialize provider session creation/deletion. Preserve upstream cleanup across client disconnects and delete sessions created after their caller disappears. Roll back driver allocation when provider registration fails, bound plugin notification requests, and match removal callbacks to their session ID. Publish provider session state atomically and add DEBUG lifecycle and Appium/WDA health diagnostics without restarting the provider.

Add regression coverage for sequential sessions, partial creation failure, client cancellation, delayed callbacks, and failed cleanup. Document the v5.7.0-to-v6.0 regression and remaining physical iOS validation.

Validation: affected Go package tests and five plugin lifecycle tests pass on develop. Race tests and go vet passed during implementation. The full Go suite requires generated GADS/docs and MongoDB for auth integration tests.

Fixes #355


docs/issue-355.md
shamanec, line 1: Please do not commit AI .md files

hub/router/appiumgrid.go
```diff
@@ -921,7 +939,7 @@ func findAvailableDevice(candidate gridCandidate, allowedWorkspaceIDs []string,
 
 			if osVersion == candidate.PlatformVersion {
 				device.Mu.Lock()
-				if device.IsAvailableForAutomation {
+				if device.IsAvailableForAutomation && !device.SessionCleanupInProgress {
```
shamanec: This is a bit weird, if `device.SessionCleanupInProgress` is true shouldn't the device not be available for automation in general?

provider/devices/appium.go
```diff
@@ -92,15 +92,18 @@ func startAppium(d PlatformDevice, capabilities models.AppiumServerCapabilities)
 		"--relaxed-security",
 		"--default-capabilities", string(capabilitiesJson))
 
-	logger.ProviderLogger.LogDebugf("device_setup", "Starting Appium on device `%s` with command `%s`", udid, cmd.Args)
+	logger.ProviderLogger.LogDebugf("appium_lifecycle", "Starting Appium udid=%s port=%s", udid, appiumPort)
```
shamanec: This log intentionally printed the full Appium startup command to check for errors, why did you remove it?

shamanec:
I am a bit skeptical about some of the changes but I don't have the time to look into the problem atm and I also don't have actual proper Appium test suites to test this so 🤷‍♂️ it seems acceptable :D

docs/issue-355.md
bjelac: Removed from the PR.

hub/router/appiumgrid.go
bjelac: Yes, you’re right. I updated the release and failed-creation paths to preserve the device’s reservation while cleanup is running. Device selection now relies on IsAvailableForAutomation
  alone. The regression test verifies that concurrent release callbacks cannot make the device available before cleanup completes.

provider/devices/appium.go
bjelac: I restored the original DEBUG log exactly as it was.

bjelac:
I created a v6.0.0-test.2 test tag in my fork and built gads-linux.zip from it. I’ll test it in our corporate environment with real Appium test suites and some actual load.

I’ll keep you posted on how it behaves and report back if I run into any issues.

shamanec:
@bjelac Okay, let me know how it goes and we can release it if it seems okay 🤞

bjelac:
Sounds good 👍

So far the fix is working well in our environment. I ran three independent test sets, each with around 10–20 Appium tests, using v6.0.0-test.2 for both the Hub and Provider.

Setup:

- Hub running as a Docker container
- Provider running natively on the Mac mini
- Real devices and real test suites in our corporate environment

All three test runs completed without any visible side effects or regressions related to the fix.

I’ll continue running it with our regular test suites and let you know if I notice any issues. If it keeps behaving like this, I’d be comfortable with releasing it.

